PR #6, opened 2017-01-10
OHFJIRA-4: admin gui redesign - integration React console

hanusto:
## Laying the foundation stone of redesigned admin GUI

Created new maven module `admin-console-client` as module that holds all web resources of OpenHub admin console, see [description](https://github.com/OpenWiseSolutions/openhub-framework/tree/feature/OHFJIRA-4-admin-gui-redesign/admin-console).

A build is realised with [frontend-plugin](https://github.com/eirslett/frontend-maven-plugin). Structure of JS app is created under root path. Result of webpack build (content `/dist` folder) is pickuped to JAR (webjar).

Location of resources is created with regard to scalability and customisation of build - /META-INF/resources/console/. This path is reserved for product-based web admin console static assets. These static assets are automatically handled by Spring based upon configuration: 

```
spring.mvc.static-path-pattern=/console/v2/**
spring.resources.static-locations=classpath:/META-INF/resources/console/
```

Actually develop URL is: `<FQHN>/admin/web/admin/console/v2/` (forwarded to `<FQHN>/admin/web/admin/console/v2/#/`). Welcome page is resolved from `spring.mvc.static-path-pattern`. I created improvement proposal for Spring Boot how welcome page can be resolved by Spring Boot itself (https://github.com/spring-projects/spring-boot/issues/7938).

From frontend point of view added new devstack - React, Redux, Mock-server, Karma, Webpack build system.


pjuza (commented):
I have two questions:
1) admin-console-client is new Maven module and this module is added into web-admin module. Why client module is not sub-module of web-admin module? What is the target architecture?

2) admin-console-client building is now part of Maven application building. I don't think it's necessary to build client everytime I need to build application - I often want to have routing core itself. I would introduce profile property ...

pjuza:
Please add/update documentation on wiki.

hanusto:
@pjuza 

ad1) web-admin will be refactored as well (see [OHFJIRA-22](https://openhubframework.atlassian.net/browse/OHFJIRA-22)). We have to create new module that represents only war build (without application code). After that this module can be moved when new structure of web-admin will be created. It is ok? I will add new TODO note for this issue.

ad2) web-admin is now part of build too, without maven profile. It is real requirement to run OpenHub without admin console? This PR is about initial support of new admin console, this is not final solution.

ad3) yes, after final solution where new created module will be and how will works.

For now, it is ok or you want to do following changes?

pjuza:
OK, I will approve this PR.

I would appreciate to build OHF core without client because I won't often require admin console to use.

pjuza (approved):

hanusto:
@pjuza ok, I will add maven profile to configure build. Default is build with or without? What is the recommended

pjuza:
Good question - default should be all-in-one approach, so default with client. And If somebody would like to turn it off then can use special profile.

hanusto:
Ok, I will do it before merge.

hanusto:
I added two maven profiles:

1. `full-build` - profile that represents full build of OHF - all modules in one. This profile is **activated by default**.
2. `full-clean` - profile that represents full clean of OHF build. It can be deactivated for example if cached admin-console libs (npm packages and so on) should be used (=> faster build).  This profile is **activated by default**.

These new maven profiles are described in [WIKI](https://openhubframework.atlassian.net/wiki/display/OHF/Maven+and+Spring).

PR #11, opened 2017-01-18
Migration all README files to markdown syntax.

hanusto:
Some README files were improved (links to [documentation](https://openhubframework.atlassian.net/wiki/display/OHF/OpenHub+framework)), for example [example-module documentation](https://github.com/OpenWiseSolutions/openhub-framework/blob/c896e4f3cf04bf488491a0d2342f27b506d3bb59/examples/README.md).

## Question
Maybe we should to avoid duplication of documentation (we have confluence, but some notes are added into README, but we have to maintain two places.


pjuza (approved):

hanusto:
@pjuza maybe you didn't see my question. 

> Maybe we should to avoid duplication of documentation (we have confluence, but some notes are added into README, but we have to maintain two places.

What do you think about that?

pjuza:
I see it's duplication but I don't see problem here because it's not huge content and I hope it won't be changing very often. 
Or how to solve it anyway? To add links to our wiki on README files?

hanusto:
Yes, for example we can add something similar: 

Check [which maven profiles](https://openhubframework.atlassian.net/wiki/display/OHF/Maven+and+Spring#MavenandSpring-Mavenprofiles) you can use and for what.

hanusto:
Ok for now I merge it and we will see if this duplication will be problem or not.

PR #17, opened 2017-01-21
Added Authentication

bionaut:
- new mock server
- login mocks
- login modal and form


pjuza (approved):

admin-console/src/common/actions/auth.actions.js
```diff
@@ -0,0 +1,18 @@
...
+export const SUBMIT_LOGIN = 'SUBMIT_LOGIN'
+
+export function toggleLoginModal () {
+  return {
+    type: LOGIN_TOGGLE
+  }
+}
+
+export function submitLogin ({ username, password }) {
+  // todo API constants ?
```
hanusto: Please could you change content type to `application/x-www-form-urlencoded`? Because now we want to support current and new console together therefore we have to use API without change.

Following snippet works correctly:
```
  // we have to use application/x-www-form-urlencoded
  const params = new URLSearchParams()
  params.append('username', username)
  params.append('password', password)
  // todo API constants ?
  const payload = axios.post('/login', params)
  return {
    type: SUBMIT_LOGIN,
    payload
  }
```

To logout, it is possible via `/logout` URL. Just call it.

When new admin console will be completely done, authentication process will use OAuth with [self-hosted OAuth2 Authorization Server](https://spring.io/guides/tutorials/spring-boot-oauth2/).

hanusto (changes requested):
Please only one change is necessary:

Rewrite `#submitLogin` function:

```
export function submitLogin ({ username, password }) {
  // we have to use application/x-www-form-urlencoded
  const params = new URLSearchParams()
  params.append('username', username)
  params.append('password', password)
  // todo API constants ?
  const payload = axios.post('/login', params)
  return {
    type: SUBMIT_LOGIN,
    payload
  }
}
```

hanusto (approved):

PR #18, opened 2017-01-23
Refine configuration model

pjuza:
This task:

- adds file _openhub.properties_ that overrides all other property files with configuration values
- removes old *.cfg files - now we use Spring Boot config model only

Look at wiki for more details: https://openhubframework.atlassian.net/wiki/display/OHF/External+configuration+model

Issue: [OHFJIRA-26](https://openhubframework.atlassian.net/browse/OHFJIRA-26)


components/src/test/java/org/openhubframework/openhub/component/ComponentTestConfig.java
```diff
@@ -16,12 +16,18 @@
...
+
+
 /**
  * Configuration for tests in "components" module.
  *
  * @author Petr Juza
  * @since 2.0
  */
+@ComponentScan(basePackages = {"org.openhubframework.openhub.component"})
+@PropertySource(value = {"classpath:/config/application-test-default.properties"})
```
hanusto, line 30: This is because since a property file application-test represents configuration of application with activated spring `test` profile, but this property file already exists, right? Maybe we can add some comment...

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Constants of core property names.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public class CoreProps {
+
+    /**
+     * FailedCount of partly fails before message will be marked as completely FAILED.
```
hanusto: `FailedCount` -> Filed count

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
```diff
@@ -0,0 +1,146 @@
...
+     * Interval (in seconds) between two tries of failed confirmations.
+     */
+    public static final String ASYNCH_CONFIRMATION_INTERVAL_SEC = "ohf.asynch.confirmation.intervalSec";
+
+    /**
+     * Skip asynchronous external calls for operation URIs that match the specified RegEx pattern.
+     */
+    public static final String ASYNCH_EXTERNAL_CALL_SKIP_URI_PATTERN = "ohf.asynch.externalCall.skipUriPattern";
+
+
```
hanusto: Remote unnecessary line please or it is intention?

core-api/src/main/java/org/openhubframework/openhub/api/asynch/AsynchConstants.java
```diff
@@ -159,7 +161,7 @@
      * URI for asynchronous message processing (with SEDA).
      */
     public static final String URI_ASYNC_MSG = "seda:asynch_message_route"
-            + "?concurrentConsumers={{ohf.asynch.concurrentConsumers}}&waitForTaskToComplete=Never"
+            + "?concurrentConsumers={{" + ASYNCH_CONCURRENT_CONSUMERS + "}}&waitForTaskToComplete=Never"
```
hanusto, line 164: Are you sure that this is correct? When camel context and `seda` is initialized, change of `ASYNCH_CONCURRENT_CONSUMERS` will not take into account. Maybe we should remove this property from DB configuration layer into "only" properties layer. Or is there another way how we can dynamically change "size" of seda?

core/src/main/java/org/openhubframework/openhub/core/alerts/AlertsPropertiesConfiguration.java
```diff
@@ -84,7 +85,7 @@ void initProps() {
         // get relevant properties for alerts
         List<String> propNames = new ArrayList<String>();
 
-        for (String propName : Tools.getAllKnownPropertyNames(env)) {
+        for (String propName : Tools.getAllKnownPropertyNames((ConfigurableEnvironment)env)) {
```
hanusto: Why casting is used? Maybe we should autowire a type safe interface.

core/src/main/java/org/openhubframework/openhub/core/common/directcall/DirectCallHttpImpl.java
```diff
@@ -47,23 +47,22 @@
...
 
     @Override
     public String makeCall(String callId) throws IOException {
         Assert.hasText(callId, "callId must not be empty");
+        Assert.hasText(localhostUri.getValue(), "localhostUri must not be empty");
 
-        CloseableHttpClient httpClient = HttpClients.createDefault();
-
-        try {
+        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
hanusto, line 59: 👍

core/src/main/java/org/openhubframework/openhub/core/common/directcall/DirectCallHttpImpl.java
```diff
@@ -47,23 +47,22 @@
...
             HttpGet httpGet = new HttpGet(localhostUri.getValue() + HTTP_URI_PREFIX
                     + DirectCallWsRoute.SERVLET_URL + "?" + DirectCallWsRoute.CALL_ID_HEADER + "=" + callId);
 
             // Create a custom response handler
             ResponseHandler<String> responseHandler = new ResponseHandler<String>() {
 
-                public String handleResponse(HttpResponse response) throws ClientProtocolException, IOException {
+                public String handleResponse(HttpResponse response) throws IOException {
                     int status = response.getStatusLine().getStatusCode();
                     if (status >= 200 && status < 300) {
```
hanusto: Maybe we can use HttpStatus enum - `org.springframework.http.HttpStatus`

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Constants of core property names.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public class CoreProps {
+
+    /**
```
hanusto: Maybe one extra PREFIX_CONSTANT can be added for reuse:
```
public static final String PREFIX = "ohf.";
```

core/src/main/java/org/openhubframework/openhub/core/confcheck/ConfigurationChecker.java
```diff
@@ -130,9 +137,14 @@ void checkConfiguration(ApplicationContext context) {
     }
 
     /**
-     * Checks if configuration parameter "ohf.contextCall.localhostUri" is valid - calls PING service.
+     * Checks if configuration parameter "ohf.server.localhostUri" is valid - calls PING service.
```
hanusto, line 140: In cluster environment, which URL will be used?

core/src/main/java/org/openhubframework/openhub/core/configuration/FileExternalPropertiesAutoConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+/**
+ * Configure and hook {@link PropertySource} from classpath:/openhub.properties external location
+ * that will override all application properties.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public class FileExternalPropertiesAutoConfiguration
```
hanusto, line 42: I see one problem of this solution - `spring.profiles.active` property is always resolved previously, so via openhub.properties there is probably no way how to setup custom spring profiles. When user of OHF needs to define own spring profiles he has to use `/config/application.properties` or JVM arguments. The same is for `spring.profiles.include`. Or maybe solution is to change behaviour of `org.springframework.boot.context.config.ConfigFileApplicationListener#addPropertySources`.

core/src/test/java/org/openhubframework/openhub/core/configuration/FileExternalPropertiesAutoConfigurationTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.openhubframework.openhub.core.AbstractCoreDbTest;
+
+
+/**
+ * Test suites that validates {@link FileExternalPropertiesAutoConfiguration}.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public class FileExternalPropertiesAutoConfigurationTest extends AbstractCoreDbTest {
```
hanusto, line 36: I think that this UC is not testing `FileExternalPropertiesAutoConfigurationTest`, it tests only external configuration model.

hanusto: Sorry, i have seen core/src/test/resources/openhub.properties :)

hanusto (changes requested):
I reviewed this PR. I think all is good, only small improvements to do added. Maybe one question is different - what is recommended way to define custom spring profiles? By current implementation of loading `openhub.properties` it is not possible.

components/src/test/java/org/openhubframework/openhub/component/ComponentTestConfig.java
pjuza: application-test-default.properties doesn' have anything with "test" profile. I used this name to describe that the file contains default values for tests ...

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
pjuza: OK, I will do it.

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
pjuza: Corrected

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
pjuza: I was intention firstly but then I forgot about it :)

core-api/src/main/java/org/openhubframework/openhub/api/asynch/AsynchConstants.java
pjuza: I only introduced constants for parameters without tryping to modify it. And when I introduced @ConfigurableValue then I haven't changed use of parameters in the code too => it's for another task.
I would mention my text from wiki:

> If you need to have one place for adding/administrating a configuration parameter common for all cluster nodes then database has to be used. Furthermore you can change parameter values on the fly (if implementation that use specific parameter can get current value anytime the parameter is used).

core/src/main/java/org/openhubframework/openhub/core/alerts/AlertsPropertiesConfiguration.java
pjuza: You are right, same with ThrottlingPropertiesConfiguration

core/src/main/java/org/openhubframework/openhub/core/common/directcall/DirectCallHttpImpl.java
pjuza: OK, I will use it ...

core/src/main/java/org/openhubframework/openhub/core/confcheck/ConfigurationChecker.java
pjuza: Each node will have its own URI (more nodes can have identical localhost URIs) - these URIs are used by nodes themselves, not for calls via load balancer

core/src/main/java/org/openhubframework/openhub/core/configuration/FileExternalPropertiesAutoConfiguration.java
pjuza: I create task https://openhubframework.atlassian.net/browse/OHFJIRA-39 for solving it in the future.
There are another possibilities how to do it at this moment, it's not critical to solve it here (it's question if it is possible):

> Command line arguments.
Properties from SPRING_APPLICATION_JSON (inline JSON embedded in an environment variable or system property)
ServletConfig init parameters.
ServletContext init parameters.
JNDI attributes from java:comp/env.

pjuza:
I went through all comments and commited few changes. Please review again ...

core-api/src/main/java/org/openhubframework/openhub/api/asynch/AsynchConstants.java
hanusto: Ok it makes sense

core/src/main/java/org/openhubframework/openhub/core/confcheck/ConfigurationChecker.java
hanusto: I know and therefore I am talking about if this property name reflects usage.

core/src/main/java/org/openhubframework/openhub/core/configuration/FileExternalPropertiesAutoConfiguration.java
hanusto: Yes, I agree. It is not critical, only my hint.

hanusto (approved):

hanusto:
Only one point: `ohf.server.localhostUri` is property to hold URI of OHF instance (node). We have task [OHFJIRA-30](https://openhubframework.atlassian.net/browse/OHFJIRA-30) to hold details about each node in DB. In property file (not shared across all nodes) there will be property to hold unique identifier of node which will represent a bind bridge to data in DB.

PR #20, opened 2017-01-27
BUGFIX: Form based security configuration for admin GUI does not work as expected

hanusto:
Fixed wrong security configuration for multiple HTTP configurer adapters.
If we have to have more than one security configurer it is necessary to split it into two adapters. The same issue occurs when we will divide spring contexts into hierarchy.

See Spring [documentation](http://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#multiple-httpsecurity) or the same problem at [Stackoverflow](http://stackoverflow.com/a/27833237).

Issue: [OHFJIRA-41](https://openhubframework.atlassian.net/browse/OHFJIRA-41)

### DIFF
--- a/web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
+++ b/web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
@@ -18,6 +18,7 @@ package org.openhubframework.openhub.admin.web.config;
 
 import static org.openhubframework.openhub.api.route.RouteConstants.WEB_URI_PREFIX;
 import static org.openhubframework.openhub.api.route.RouteConstants.WS_AUTH_POLICY;
+import static org.openhubframework.openhub.api.route.RouteConstants.WS_URI_PREFIX;
 
 import java.util.Collections;
 
@@ -27,6 +28,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
 import org.springframework.security.access.AccessDecisionManager;
 import org.springframework.security.access.AccessDecisionVoter;
 import org.springframework.security.access.vote.AffirmativeBased;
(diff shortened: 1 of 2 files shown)


web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
```diff
@@ -61,63 +63,94 @@
...
+            // @formatter:on
+        }
+    }
+
+    /**
+     * Basic configuration for admin console, handled by {@link RouteConstants#WEB_URI_PREFIX}. 
+     */
+    @Configuration
+    @Order(ConsoleSecurityConfig.ORDER)
+    public static class ConsoleSecurityConfig extends WebSecurityConfigurerAdapter {
```
pjuza: When anybody says "console" then I imagine "command line console". What about AdminGuiConsole or only AdminGui ...

web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
```diff
@@ -61,63 +63,94 @@
...
+                    .sessionManagement()
+                        .maximumSessions(1)
+                        .expiredUrl(LOGIN_PAGE_URL + "?expired");
+            // @formatter:on
+        }
     }
 
-    @Override
+    @Autowired // global authentication configuration for all adapters
    	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
```
pjuza: Would be possible to override this settings in specific project? For example load users from LDAP ...

pjuza (approved):

web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
hanusto: Ok, good point. I will use admin GUI, but in future. Admin from backend point of view will be represented via REST endpoints.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/config/WebSecurityConfig.java
hanusto: Good question. It is possible to override this settings by specific project as in previous version. `AuthenticationManagerBuilder` is object exposed by `AuthenticationConfiguration` that is active if `@EnableWebSecurity` is used. The most important from extendability point of view is `WebSecurityConfigurerAdapter#init` where protected method `WebSecurityConfigurerAdapter#getHttp` is called. This method calls `WebSecurityConfigurerAdapter#authenticationManager`, where main message is:

- if authentication manager is initialized, it will be used, otherwise authentication manager created from builder is used (not exposed as bean).

So it does not matter if implementation class overrides `#configure` method or use `@Authowire` approach via `#configure`, result is the same. If project wants to override this behaviour, it is necessary to create authentication manager and expose it as bean.

BWT: currently I return back this change.

hanusto:
I changed the name of configuration class from ConsoleSecurityConfig to AdminGuiSecurityConfig. Answer to question about extendability of authentication manager was clarified. From my side it is done.

PR #23, opened 2017-02-02
OHFJIRA-40: Isolation of web/admin context

hanusto:
[Isolation of web context in Spring Boot](https://github.com/bijukunjummen/sample-spring-boot-hierarchy), specially for admin components, via [Spring Boot Application fluent builder](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-spring-application.html#boot-features-fluent-builder-api). Web context is child spring context of parent that is represented by OpenHubApplication configuration. Separation of context also reflects Spring Boot [security configuration](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-security).

## Some useful links:
- context overview: http://stackoverflow.com/questions/11708967/what-is-the-difference-between-applicationcontext-and-webapplicationcontext-in-s
- http://stackoverflow.com/questions/33739359/combining-basic-authentication-and-form-login-for-the-same-rest-api (multiple)
- multiple security context: 
	- http://stackoverflow.com/questions/1070420/can-i-have-multiple-security-contexts-with-spring-security
	- security context in child: http://stackoverflow.com/questions/16920243/how-to-have-spring-security-context-in-child-context
	- http://stackoverflow.com/questions/30670327/spring-boot-with-multiple-dispatcherservlet-each-having-their-own-controllers

The most important classes:
- `FilterSecurityInterceptor` - interceptor which is responsible for validation of incoming request based upon configuration defined via `WebSecurityConfigurerAdapter` (specially method `org.springframework.security.access.intercept.AbstractSecurityInterceptor#beforeInvocation`)

## Changes:
- added URL constants
- version printer improvement
- camel config - move servlet registration
- web service config - move servlet registration
- `OpenHubApplication` - [create parent/child context hierarchy](https://www.javacodegeeks.com/2016/12/spring-boot-application-context-hierarchy.html)
	- `WebContextConfig` as [web child context configurer](http://www.kubrynski.com/2014/01/understanding-spring-web-initialization.html), it handles `org.openhubframework.openhub.web` and `org.openhubframework.openhub.admin` packages
		- registers beans, servlets and filters
	- `WebConfigurer` - is responsible for configuration of web runtime layer of OpenHub, for example servlet registration of camel connectors.
        - `AdminMvcConfig` - added [handling of freemarker static resources](http://stackoverflow.com/questions/28725635/spring-boot-configure-it-to-find-the-webapp-folder) (from WEB-INF)
- separation of security configuration - authentication manager `GlobalSecurityConfig` and for root (`WebSecurityConfig`) and admin context (`AdminSecurityConfig`), [supported](http://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#multiple-httpsecurity) [basic auth and form based for admin](http://stackoverflow.com/a/27833237/2596932)
	- `GlobalSecurityConfig` - auto-configuration of authentication manager, it is simple possible to override this with custom implementation
	- `WebSecurityConfig` - _accessDecisionManager_ and _authorizationPolicy_ are exposed only when the specified bean classes and/or names are
  		not already contained in the _BeanFactory_.
	- `AdminSecurityConfig` - configuration of security for /web/admin paths
- new package hierarchy definition
	- root context - main configuration of OpenHub - `org.openhubframework.openhub.config`
	- child context
		- web  - `org.openhubframework.openhub.web` auto-scan
		- admin - `org.openhubframework.openhub.admin` auto-scan

**See:**
contexts
- [application.properties](https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html)
	- identity of application
	- configure live beans view for /beans actuator to reflect current parent-child hierarchy
	- management path is /mgmt
	- white label error page disable
	- deactivate `X-Application-Context` HTTP Header  
- freemarker templates - corrected login and logout handling

Issue: [OHFJIRA-40](https://openhubframework.atlassian.net/browse/OHFJIRA-40)


web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
+import org.springframework.security.config.annotation.authentication.configurers.GlobalAuthenticationConfigurerAdapter;
+
+import org.openhubframework.openhub.common.AutoConfiguration;
+
+
+/**
+ *  Global <strong>default</strong> security configuration of OpenHub.
+ *
```
pjuza: Maybe use html element p

web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
```diff
@@ -0,0 +1,73 @@
...
+        auth.inMemoryAuthentication()
+                .withUser(defaultUsers.getMonitoringUser()).password(defaultUsers.getMonitoringPassword())
+                    .roles(AuthRole.MONITORING.name());        
+        // @formatter:on
+    }
+
+    /**
+     * Default authentication roles definition.
+     */
+    public enum AuthRole {
```
pjuza, line 67: Only idea - we have fixed list of roles for OHF core. I can define my own roles in specific project, I can define security for my project and how can I say that those custom roles can have access to admin GUI or monitoring:

- isn't it wrong approach? maybe yes
- I can override global security with my custom roles => it's not for discussion anymore

web-admin/src/main/java/org/openhubframework/openhub/config/WebContextConfig.java
```diff
@@ -0,0 +1,170 @@
...
+@EnableAutoConfiguration
+@EnableConfigurationProperties
+@ComponentScan(basePackages = {
+        "org.openhubframework.openhub.web",
+        "org.openhubframework.openhub.admin"
+},
+        excludeFilters =
+        @ComponentScan.Filter(type = FilterType.ANNOTATION, classes = AutoConfiguration.class))
+@Configuration
+@ImportResource({"classpath:net/bull/javamelody/monitoring-spring.xml", "classpath:sp_h2_server.xml"})
```
pjuza, line 70: Is it correct to have them in web context when these features (h2, monitoring) are related to core module?

web-admin/src/main/java/org/openhubframework/openhub/web/WebConfigurer.java
```diff
@@ -0,0 +1,87 @@
...
+public class WebConfigurer {
+
+    /**
+     * Configures servlet for HTTP communication.
+     *
+     * @return registration bean of {@link CamelHttpTransportServlet}.
+     */
+    @Bean(name = RouteConstants.CAMEL_SERVLET)
+    @ConditionalOnMissingBean(name = RouteConstants.CAMEL_SERVLET)
+    public ServletRegistrationBean camelHttpServlet() {
```
pjuza, line 51: I'm a little bit confused because I thought camel servlet and WS dispatcher would be part of parent context, not web context? Why did you make it this way?

web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
hanusto: For what? I don't understand what I should do :). Can you please write exact change?

web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
hanusto: In my opinion recommended way is to override global security configuration and you can use out of box roles or create custom one. There are only two places (logically only one) where a role is used - for security configuration. I think that when project will want to change something it will be necessary to do more than only create custom roles.

Are you agree or I should do something?

web-admin/src/main/java/org/openhubframework/openhub/config/WebContextConfig.java
hanusto: What is core module? Because maybe this question is similar as comment above.

I think that **web** monitoring and **web** h2 console are **web-based** functions and therefore these functions should be provided via web context. Moreover Spring Boot does not allow to specify parent-child context hierarchy where parent has web components. These components are deactivated at all. 

[See documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-spring-application.html#boot-features-fluent-builder-api): There are some restrictions when creating an `ApplicationContext` hierarchy, e.g. **Web components must be contained within the child context**, and the same Environment will be used for both parent and child contexts. See the SpringApplicationBuilder Javadoc for full details.

web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
pjuza: You start new paragraph without using html element.

web-admin/src/main/java/org/openhubframework/openhub/config/GlobalSecurityConfig.java
pjuza: I agree ...

web-admin/src/main/java/org/openhubframework/openhub/web/WebConfigurer.java
hanusto: I described a reason in previous comment. Generally there are two ways how child context can be defined:

- via [SpringApplicationBuilder](http://docs.spring.io/spring-boot/docs/1.5.1.RELEASE/api/org/springframework/boot/builder/SpringApplicationBuilder.html), but generally Spring Boot has limitation for creating parent/child context hierarchy, see [source code](https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java#L190):

```
public SpringApplicationBuilder child(Object... sources) {
   SpringApplicationBuilder child = new SpringApplicationBuilder();
   child.sources(sources);

   // Copy environment stuff from parent to child
   child.properties(this.defaultProperties).environment(this.environment)
      .additionalProfiles(this.additionalProfiles);
   child.parent = this;

   // It's not possible if embedded containers are enabled to support web contexts as
   // parents because the servlets cannot be initialized at the right point in
   // lifecycle.
   web(false);

   // Probably not interested in multiple banners
   bannerMode(Banner.Mode.OFF);

   // Make sure sources get copied over
   this.application.setSources(this.sources);

   return child;
}
```

> It's not possible if embedded containers are enabled to support web contexts as
> parents because the servlets cannot be initialized at the right point in
> lifecycle.

So all web components must be defined in child context, strictly speaking child context which has no another child context.

- manually via `AnnotationConfigWebApplicationContext`:

```
@Bean
public ServletRegistrationBean adminConsoleServletRegistrationBean() {
    DispatcherServlet dispatcherServlet = new DispatcherServlet();
    AnnotationConfigWebApplicationContext context = new AnnotationConfigWebApplicationContext();
    context.setParent(rootContext);
    context.setDisplayName("Admin Application Context");
    context.setId("AdminConsoleApplicationContext");
    context.register(AdminConsoleContextConfig.class);
    dispatcherServlet.setApplicationContext(context);
    ServletRegistrationBean registration = new ServletRegistrationBean(dispatcherServlet, WEB_URI_PREFIX_MAPPING);
    registration.setName("adminDispatcherServlet");
    registration.setLoadOnStartup(1);
    return registration;
}
```

Interesting is that this choice supports web components in parent context (are not deactivated) but problem is exactly in lifecycle of parent context. For example when application is going to shutdown, camel context (participates in parent context) is graceful shutdown too, but `ContextRefreshedEvent` is fired and parent (camel) context is restarted, but in this moment, parent context (root) is already closed.

Now I don't know if definition of web-based components (in this example dispatcher servlets)  in web context is wrong. Why should be?

web-admin/src/main/java/org/openhubframework/openhub/config/WebContextConfig.java
pjuza: Thank you for clarification ...

pjuza (approved):

web-admin/src/main/java/org/openhubframework/openhub/web/WebConfigurer.java
hanusto: Maybe one more thing - `web-admin` maven module is confusing, because here is for example `OpenHubApplication`. We have task to support simple extending and using OpenHub for projects and one thing which has to be done is separation of `web-admin` module into two modules:

- web - module that is responsible for initialization of OpenHub from web point of view, but it will be jar, so all components can be extended or overridden. In war it is not possible directly (classes from war package is not visible) - I mean for example exposing beans by type, but that is all.
- war - module that represents only packaging of OpenHub

hanusto:
So are you agree that moving web components from parent context into web context is OK? By the way, what you thought about `p` html element in first comment? I don't understand that :).

pjuza:
I have already approved this PR, I agree with your clarification ...

PR #24, opened 2017-02-03
OHFJIRA-20 - Standalone (executable) OHF running

hanusto:
Configuration of build to support standalone (executable) OHF. 

openhubstandalone

To run OHF in standalone mode it is necessary:

- build OHF with `esb.executable` maven profile: `mvn clean && mvn package -P esb.executable`
- run from command line (or IDE)

```
java $JAVA_OPTS -Dserver.port=8080 \
    -jar web-admin/target/openhub.war
```
Running of OHF in standalone mode is described at [WIKI](https://openhubframework.atlassian.net/wiki/display/OHF/Standalone+running).

There are some issues which had to be fixed, for example hotfix of Freemarker configuration for embedded containers - https://github.com/spring-projects/spring-boot/issues/907. It is also necessary to include `javax.servlet:jsp-api` dependency. When OHF is started in standalone mode there is issue (it is issue of logging configuration) of Logback. Custom solution is created for lazy Logback appender initialization. It is necessary to use, because OpenHub uses condition for appenders (DEV/PROD profile) and if this appender is not active, a file is created anyway.

It is not necessary to use `ContextRefreshedEvent` in `ConfigurationChecker` but `ContextStartedEvent` is. `ConfigurationChecker` checks localhost URI, but it is necessary that web context is already fully started. When `ContextRefreshedEvent` is fired we don’t know if web context is started and this check fails.

Some H2 configuration improvements added - connection to mix mode and deactivate auto-close of connections.

## Changes:
- fixed wrong start-class property to correct Spring Boot application bootstrapping
- **added run script** `run-ohf.sh`
- configure PID and PORT lister to support run script
- added `esb.executable` maven profile to add necessary dependencies - DB drivers, connection pooler, embedded tomcat container and related libs
- package executable OHF as fat war


pjuza (approved):
Perfect, I don't have any comments ...

PR #26, opened 2017-02-06
Add memory-grid/cache

pjuza:
Overview of use cases where clustered cache/memory-grid could be used:

- throttling
- cache (static) external content (e.g. codebases, static file content, ...)
- cache internal OHF data
- cache configuration properties 
- use shared memory for saving runtime data, e.g. node evidence, evidence of running scheduled jobs
- cache for specific business logic implementations on projects

We decided to use Hazelcast memory grid. See [OHFJIRA-36](https://openhubframework.atlassian.net/browse/OHFJIRA-36) for more details why we selected Hazelcast.

Summary of the work:
- added throttling implementation by Hazelcast shared map
- added default Hazelcast configuration file "ohf_hazelcast.xml"
- added Hazelcast dependency
- added default cache/hazelcast configuration
- added new "cluster" Spring profile

Issue: [OHFJIRA-36](https://openhubframework.atlassian.net/browse/OHFJIRA-36)

### DIFF
--- a/common/src/main/java/org/openhubframework/openhub/common/OpenHubPropertyConstants.java
+++ b/common/src/main/java/org/openhubframework/openhub/common/OpenHubPropertyConstants.java
@@ -33,7 +33,7 @@ public class OpenHubPropertyConstants {
      * OpenHub property prefix name used for custom properties created for OpenHub.
      * These property names should have the same prefix.
      */
-    public static final String PREFIX = "ohf";
+    public static final String PREFIX = "ohf.";
 
     private OpenHubPropertyConstants() {
         // avoid instantiation
--- a/core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.core.throttling;
+
+import javax.annotation.Nullable;
+
+import org.apache.commons.lang3.StringUtils;
+import org.joda.time.LocalDateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.Assert;
+
+import org.openhubframework.openhub.spi.throttling.ThrottleCounter;
+import org.openhubframework.openhub.spi.throttling.ThrottleScope;
+
+
+/**
+ * Abstract implementation of {@link ThrottleCounter} interface.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public abstract class AbstractThrottleCounter implements ThrottleCounter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractThrottleCounter.class);
+
+    private static final int DUMP_PERIOD_SEC = 60;
+
+    private volatile LocalDateTime lastDumpTimestamp = LocalDateTime.now();
+
+    @Override
+    public final int count(ThrottleScope throttleScope, int intervalSec) {
+        Assert.notNull(throttleScope, "the throttleScope must not be null");
+        Assert.isTrue(intervalSec > 0, "the intervalSec must be positive value");
+
+        Integer count = doCount(throttleScope, intervalSec);
+
+        // make dump only once in the specified intervalSec
+        if (LOG.isDebugEnabled() && (LocalDateTime.now().minusSeconds(DUMP_PERIOD_SEC).isAfter(lastDumpTimestamp))) {
+            String cacheInfo = getCacheInfo();
+
+            if (StringUtils.isNotEmpty(cacheInfo)) {
+                LOG.debug(cacheInfo);
+            }
+
+            lastDumpTimestamp = LocalDateTime.now();
+        }
+
+        return count;
+    }
+
+    /**
+     * Counts requests for specified throttle scope and time interval.
+     *
+     * @param throttleScope the throttle scope
+     * @param intervalSec the time interval in seconds
+     * @return count of requests
+     */
+    protected abstract int doCount(ThrottleScope throttleScope, int intervalSec);
+
+    /**
+     * Returns memory info/statistics for logging.
+     *
+     * @return cache info or {@code null} if there is no info to log
+     */
+    @Nullable
+    String getCacheInfo() {
+        return null;
+    }
+}
--- a/core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.core.throttling;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Stack;
+import javax.annotation.Nullable;
+
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
+import com.hazelcast.map.AbstractEntryProcessor;
+import com.hazelcast.map.EntryProcessor;
+import org.joda.time.DateTime;
+import org.springframework.util.Assert;
+
+import org.openhubframework.openhub.spi.throttling.ThrottleCounter;
+import org.openhubframework.openhub.spi.throttling.ThrottleScope;
+
+
+/**
+ * Shared memory (aka memory grid or cache) implementation of {@link ThrottleCounter} interface
+ * by Hazelcast' {@link IMap map} and {@link EntryProcessor}.
+ * Suitable for running OpenHub in the cluster environment.
+ * <p/>
+ * Implementation prerequisites Hazelcast's map definition with the name '{@value #MAP_THROTTLING}'.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ * @see HazelcastThrottleScope
+ */
+public class ThrottleCounterHazelcastImpl extends AbstractThrottleCounter {
+
+    private static final String MAP_THROTTLING = "throttling";
+
+    private final HazelcastInstance hazelcast;
+
+    public ThrottleCounterHazelcastImpl(HazelcastInstance hazelcast) {
+        Assert.notNull(hazelcast, "hazelcast must not be null");
+
+        this.hazelcast = hazelcast;
+    }
+
+    @Override
+    protected int doCount(ThrottleScope throttleScope, int intervalSec) {
+        IMap<HazelcastThrottleScope, List<Long>> map = hazelcast.getMap(MAP_THROTTLING);
+
+        Assert.notNull(map, "shared map must not be null");
+
+        HazelcastThrottleScope sharedThrottleScope = new HazelcastThrottleScope(throttleScope);
+        return (Integer) map.executeOnKey(sharedThrottleScope, new CounterEntryProcessor(intervalSec));
+    }
+
+    @Override
+    @Nullable
+    String getCacheInfo() {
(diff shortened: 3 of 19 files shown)


core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
```diff
@@ -0,0 +1,146 @@
...
+ * <p/>
+ * Implementation prerequisites Hazelcast's map definition with the name '{@value #MAP_THROTTLING}'.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ * @see HazelcastThrottleScope
+ */
+@Profile(Profiles.CLUSTER)
+@Service
+@Primary
```
hanusto: Why this implementation is marked as `@Primary`? How project makes primary bean of primary bean? Maybe this implementation should be exposed as configuration approach:

1) if profile is cluster
2) if classpath contains hazelcast
3) if no other AbstractThrottleCounter based implementation bean does exist

-> this implementation is exposed. I am talking about `OpenHubCacheAutoconfiguration` class. Next advantage is that project can directly drive order of instantiation.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
```diff
@@ -0,0 +1,146 @@
...
+public class ThrottleCounterHazelcastImpl extends AbstractThrottleCounter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ThrottleCounterHazelcastImpl.class);
+
+    private static final String MAP_THROTTLING = "throttling";
```
hanusto: It is similar as region?

core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
```diff
@@ -0,0 +1,74 @@
...
+     * Counts requests for specified throttle scope and time interval.
+     *
+     * @param throttleScope the throttle scope
+     * @param intervalSec the time interval in seconds
+     * @return count of requests
+     */
+    protected abstract int doCount(ThrottleScope throttleScope, int intervalSec);
+
+    /**
+     * Dumps memory or memory statistics to log on "debug" level.
```
hanusto: I think that what `#dumpMemory()` method do is implementation detail. Certainly log level.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
hanusto: Why this implementation is marked as @Primary? How project makes primary bean of primary bean? Maybe this implementation should be exposed as configuration approach:

- if profile is cluster
- if classpath contains hazelcast
- if no other AbstractThrottleCounter based implementation bean does exist

-> this implementation is exposed. I am talking about `OpenHubCacheAutoconfiguration` (or similar) class. Next advantage is that project can directly drive order of instantiation.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    protected int doCount(ThrottleScope throttleScope, int intervalSec) {
+        IMap<HazelcastThrottleScope, List<Long>> map = hazelcast.getMap(MAP_THROTTLING);
+
+        Assert.notNull(map, "shared map must not be null");
+
+        HazelcastThrottleScope sharedThrottleScope = new HazelcastThrottleScope(throttleScope);
+        return (Integer) map.executeOnKey(sharedThrottleScope, new CounterEntryProcessor(intervalSec));
```
hanusto, line 65: Perfect, calculates throttling count in datagrid. 👍

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
```diff
@@ -0,0 +1,146 @@
...
+    @Autowired
+    public ThrottleCounterHazelcastImpl(HazelcastInstance hazelcast) {
+        this.hazelcast = hazelcast;
+    }
+
+    @Override
+    protected int doCount(ThrottleScope throttleScope, int intervalSec) {
+        IMap<HazelcastThrottleScope, List<Long>> map = hazelcast.getMap(MAP_THROTTLING);
+
+        Assert.notNull(map, "shared map must not be null");
```
hanusto, line 62: Maybe if this does not exist, we can initialize it instead of throw exception.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterMemoryImpl.java
```diff
@@ -42,27 +40,20 @@
  * @author Petr Juza
  */
 @Service
-public class ThrottleCounterMemoryImpl implements ThrottleCounter {
+public class ThrottleCounterMemoryImpl extends AbstractThrottleCounter {
```
hanusto: I think that this implementation should be exposed via autoconfiguration approach.

core/src/main/resources/config/ohf_hazelcast.xml
```diff
@@ -0,0 +1,144 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
```
hanusto: This configuration is our, it should be 2017

hanusto (commented):
Globally it looks like perfect but I have question how default implementation of throttling counter is initialized. I think that we should prepare auto-configuration class which will be responsible for initializing of specific bean based upon conditions. For example default implementation is InMemory, but if `cluster` spring profile is active and also hazelcast classes are in classpath, `ThrottleCounterHazelcastImpl` is prefered. I think that `@Primary` annotation is dangerous and should be used only by projects which want to override product defaults.

core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
pjuza: I have changed the method - dumpMethod -> getCacheStatistics that returns String for loggind at one place in DEBUG level.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
pjuza: Hazelcast's map with the name "throttling" is defined in "ohf_hazelcast.xml". It's prerequisite to have hazelcast configuration - either this default or custom (defined by spring.hazelcast.config property). 
Map definition is in XML format by design because it will be changed for specific cluster deployment.

core/src/main/resources/config/ohf_hazelcast.xml
pjuza: I have changed it but it's question if this licence is our (because it's our custom configuration) or Hazelcast's licence because it's theirs XSD.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
pjuza: It's name of the map specific definition for throttling, see "ohf_hazelcast.xml"

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
pjuza: I created org.openhubframework.openhub.core.throttling.ThrottlingAutoConfiguration for better configuration.

pjuza:
I have commited my refinements, please review it again.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
hanusto: 👍

common/src/main/java/org/openhubframework/openhub/common/OpenHubPropertyConstants.java
```diff
@@ -29,12 +29,6 @@
      */
     public static final String SPRING_PREFIX = "spring";
 
-    /**
-     * OpenHub property prefix name used for custom properties created for OpenHub.
-     * These property names should have the same prefix.
-     */
-    public static final String PREFIX = "ohf";
```
hanusto: Why this prefix is deleted? We should use it to avoid collision among product and project implementations. I see this in `org.openhubframework.openhub.api.configuration.CoreProps#PREFIX`. It is correct place I would say 👍

core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    protected abstract int doCount(ThrottleScope throttleScope, int intervalSec);
+
+    /**
+     * Returns memory info/statistics for logging.
+     *
+     * @return cache info or {@code null} if there is no info to log
+     */
+    @Nullable
+    String getCacheInfo() {
```
hanusto, line 81: 👍

core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
```diff
@@ -0,0 +1,84 @@
...
+ * Abstract implementation of {@link ThrottleCounter} interface.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+public abstract class AbstractThrottleCounter implements ThrottleCounter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractThrottleCounter.class);
+
+    private static final int DUMP_PERIOD_SEC = 60;
```
hanusto, line 41: Maybe one improvement can be add possibility to define custom period of dumps. But is is not necessary, only idea...

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottleCounterHazelcastImpl.java
hanusto: Ok, thanks for clarification.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottlingAutoConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Configures throttling counter implementations.
+ * There is property '{@value #COUNTER_IMPL_PROPERTY}' that defines which built-in implementation will be used.
+ * If not defined then default {@link ThrottleCounterMemoryImpl} implementation is used.
+ *
+ * @author Petr Juza
+ * @since 2.0
+ */
+@Configuration
```
hanusto: Why `@AutoConfiguration` is not used instead? In OpenHub `@AutoConfiguration` has strict lifecycle driven by Spring Boot via `spring.factory`, `@Configuration` has problem for example in tests. I recommend to change it.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottlingAutoConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+@ConditionalOnMissingBean(ThrottleCounter.class)
+@EnableConfigurationProperties(ThrottlingProperties.class)
+public class ThrottlingAutoConfiguration {
+
+    private static final String COUNTER_IMPL_PROPERTY = "ohf.throttling.counter.impl";
+
+    private static final String IN_MEMORY_CLASS_NAME
```
hanusto, line 48: Maybe `IN_MEMORY_CLASS_NAME = ThrottleCounterMemoryImpl.class.getCanonicalName();` and for `HAZELCAST_CLASS_NAME` the same, it is safe for refactoring for example.

hanusto (approved):
Now auto-configuration looks like perfect 👍.

I approve this PR and recommend to consider following:
- `ThrottlingAutoConfiguration` should have `@AutoConfiguration` annotation - this annotation is excluded from auto-scan (see `OpenHubApplication`). 
- improvement how `IN_MEMORY_CLASS_NAME` and `HAZELCAST_CLASS_NAME` are defined.

hanusto:
By the way, documentation should be updated.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottlingAutoConfiguration.java
pjuza: I have change it.

core/src/main/java/org/openhubframework/openhub/core/throttling/ThrottlingAutoConfiguration.java
pjuza: I wanted to define this way but I got compilation error from @ConditionalOnProperty, attribute havingValue: "Error:(53, 95) java: element value must be a constant expression"
I don't know how to do it in another way?

hanusto: Ok, it makes sense, sorry.

common/src/main/java/org/openhubframework/openhub/common/OpenHubPropertyConstants.java
pjuza: There was duplication because the same constant was also in org.openhubframework.openhub.api.configuration.CoreProps.
I changed it back and removed from CoreProps ...

core/src/main/java/org/openhubframework/openhub/core/throttling/AbstractThrottleCounter.java
pjuza: I have been thinking about it but then I decided not to change it because it's only for debugging and main configuration parameter is logging level.

PR #27, opened 2017-02-08
Stabilization and unification of logging system

hanusto:
Default logging platform of OpenHub is [logback as for Spring Boot itself](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-logging.html) but it can be exchanged. We want to support configuration of logging via props, so there is bridge between logback configuration and property values. Configuration of logback is changed because it should represent production-ready configuration (avoid DEBUG logging and so on). On the other hand there is new application-dev.properties which increase logging level (to DEBUG) if spring profile DEV is active. Logback configuration is separated into **ohf-logback.xml** that is included into **logback-spring.xml** to support simple reuse of core logback configuration by projects (the same approach as for Spring Boot, see for example _org/springframework/boot/logging/logback/defaults.xml_). Spring recommends to use `-spring` variants for your logging configuration (for example `logback-spring.xml` rather than `logback.xml`). If we use standard configuration locations, Spring cannot completely control log initialization.

Since logging is initialized before the ApplicationContext is created, it isn’t possible to control logging from `@PropertySources` in Spring @Configuration files. System properties and the conventional Spring Boot external configuration files work just fine.).

### Some useful links:

- https://springframework.guru/using-logback-spring-boot/
- root level: http://stackoverflow.com/questions/20485059/spring-boot-how-can-i-set-the-logging-level-with-application-properties

Documentation will be updated ASAP when we will have consensus if this setup is correct.

## Changes:
- separation of `logback-spring.xml`/ `ohf-logback.xml` (recommended name)
- new application-dev.properties (removed DEV profile as default activated profile)
- improved console and file pattern (indent of messages - simple parsing log files, ANSI colour)
- bridge between property values and logback configuration
- improved configuration of JPA (hibernate) logging - DEBUG only for dev spring profile
- configured logging platform in tests.
- Logging filter is used only for administration endpoints to avoid duplication of log events, for example logging XML requests and responses twice.

Issue: [OHFJIRA-21](https://openhubframework.atlassian.net/browse/OHFJIRA-21)


pjuza (approved):
I have no comments, it's ok.

PR #28, opened 2017-02-08
Nodes evidence for cluster environment

openhubframework:
In Openhub we need register all nodes in cluster environment. 
**Main function:**

- information about nodes are saved in database
- when infomation about node is not found in database, it will be crated when the application server starts
- every node can be in three states:
  * _RUN_ - Node handles new and existing (saved messages in ESB) messages
  * _HANDLES_EXISTING_MESSAGES_ - Node handles only existing messages. New messages/requests are rejected
  * _STOPPED_ - Node not handles existing messages and new messages
- on message entity is information which node processing message
- in configuration with key _ohf.cluster.actualNodeInstance.code_ is saved, which node from database is actual node of application server

**Jira task**
https://openhubframework.atlassian.net/browse/OHFJIRA-30


pjuza:
Firstly add description to this PR to describe what you are solving and how it is solved.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
```diff
@@ -813,6 +834,7 @@ public String toString() {
             .append("guaranteedOrder", guaranteedOrder)
             .append("excludeFailedState", excludeFailedState)
             .append("processingPriority", processingPriority)
```
pjuza, line 834: Is correct to have guaranteedOrder / excludeFailedState / processingPriority in equals method?

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * @author Roman Havlicek
+ * @since 2.0.0
+ */
+@Entity
+@Table(name = "node",
+        uniqueConstraints = {
+                @UniqueConstraint(name = "uq_node_code", columnNames = {"code"}),
+                @UniqueConstraint(name = "uq_node_name", columnNames = {"name"})})
+public class Node implements HumanReadable {
```
pjuza: Wouldn't be better to use "cluster_node" instead of node. When somebody says node then I imagine XML node firstly ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
```diff
@@ -0,0 +1,229 @@
...
+                .toHashCode();
+    }
+
+    @Override
+    public String toString() {
+        return new ToStringBuilder(this)
+                .append("nodeId", nodeId)
+                .append("code", code)
+                .append("name", name)
+                .append("description", description)
```
pjuza: Use trim() because description can have over 2000 characters ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
```diff
@@ -0,0 +1,229 @@
...
+    public String getCode() {
+        return code;
+    }
+
+    /**
+     * Sets code of this node.
+     *
+     * @param code code
+     */
+    public void setCode(String code) {
```
pjuza: Is it ok that code/name can be set when Node is created? When node is created why somebody want's to change code or name?
Hibernate (JPA) doesn't need to have public set* methods.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
```diff
@@ -0,0 +1,229 @@
...
+    public NodeState getState() {
+        return state;
+    }
+
+    /**
+     * Sets state of this node.
+     *
+     * @param state state
+     */
+    public void setState(NodeState state) {
```
pjuza: I would prefer to have as much business logic as possible in true objects, not in services. Avoid to anemic domain model (https://en.wikipedia.org/wiki/Anemic_domain_model).

I suggest the following:
- replace setState with methods such as run(), stop(), ...
- add business logic checks if transfers between different states are possible

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Gets if {@link Node} in this state process messages existing saved in ESB (from SEDA or
+     * {@link MsgStateEnum#PARTLY_FAILED}, {@link MsgStateEnum#POSTPONED}, etc.).
+     *
+     * @return {@code true} node process existing saved messages, {@code false} - otherwise
+     */
+    public boolean isProcessExistingMessages() {
+        return processExistingMessages;
+    }
+}
```
pjuza: Please add toString()

core-spi/src/main/java/org/openhubframework/openhub/spi/node/NodeService.java
```diff
@@ -0,0 +1,132 @@
...
+ * @since 2.0.0
+ */
+public interface NodeService {
+
+    /**
+     * Insert new {@link Node}.
+     *
+     * @param node node
+     */
+    void insertNode(Node node);
```
pjuza: I know it's not for thisissue but I would like to change like in WisePorter - to have true objects which will be able to saves into DB themselves. Not to have services with methods which manipulate with object's state.

core-spi/src/main/java/org/openhubframework/openhub/spi/node/NodeService.java
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param node node on which will be set state
+     */
+    void setStoppedState(Node node);
+
+    /**
+     * Set on actual node state when node is stopped (new message will be rejected and existing is not processing)
+     * {@link NodeState#STOPPED}.
+     */
+    void setStoppedStateOnActualNode();
```
pjuza: You should move all these methods to Node object, wouldn't it be better?
If it's here because of transaction demarcation then we should consider using AOP (https://openhubframework.atlassian.net/browse/OHFJIRA-32)

core-spi/src/main/java/org/openhubframework/openhub/spi/node/NodeService.java
```diff
@@ -0,0 +1,132 @@
...
+     * @return node for this server instance
+     */
+    Node getActualNode();
+
+    /**
+     * Get actual node identifier for this application server instance.
+     *
+     * @return node identifier for this server instance
+     */
+    Long getActualNodeId();
```
pjuza: Why getActualNodeId when you have getActualNode ?

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+public class NodeServiceImpl implements NodeService, InitializingBean {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NodeServiceImpl.class);
+
+    private final TransactionTemplate transactionTemplate;
+
+    @Autowired
+    private NodeDao nodeDao;
+
+    @Value("${node.actualInstance.code}")
```
pjuza: Use constant for property name + add prefix org.openhubframework.openhub.common.OpenHubPropertyConstants#PREFIX

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+        Assert.notNull(node.getNodeId(), "only new node can be delete");
+
+        nodeDao.delete(node);
+
+        LOG.debug("Deleted node " + node.toHumanString());
+    }
+
+    @Override
+    public List<Node> getAllNodes() {
+        return nodeDao.getAllNodes();
```
pjuza: I would use Collections.unmodified ...

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+    public Long getActualNodeId() {
+        if (actualNodeId == null) {
+            throw new IllegalStateException("Actual node id is not defined.");
+        }
+        return actualNodeId;
+    }
+
+    @Override
+    public boolean isActualNodeStopping() {
+        return getActualNode().getState().equals(NodeState.STOPPED);
```
pjuza: I prefer to move all these methods to Node object, encapsulate logic around node's state

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+    @Override
+    @Transactional
+    public void setStoppedStateOnActualNode() {
+        setStoppedState(getActualNode());
+    }
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        //test if configuration value exist
+        if (StringUtils.isBlank(actualNodeCode)) {
```
pjuza: Please implement org.openhubframework.openhub.core.confcheck.ConfCheck contract
See org.openhubframework.openhub.core.confcheck.ConfigurationChecker

core/src/main/resources/db/db_schema_postgreSql_2_0.sql
```diff
@@ -32,4 +32,38 @@ CREATE INDEX configuration_item_cat_code_idx ON configuration_item (category_cod
...
+--
+
+drop table if exists node cascade;
+
+create table node (
+    node_id         int8            not null,
+    code            varchar(256)    not null unique,
+    name            varchar(256)    not null unique,
+    description     varchar(2056)   null,
+    state           varchar(64)     not null,
```
pjuza: Isn't it too much for enum item?

pjuza (changes requested):
Main points from my perspective:
- encapsulate as much as possible logic into Node object (I know that Message is not the right example but we need to refactor it in the meantime.
- use ConfCheck

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
openhubframework: Why toString() method in ENUM?

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
openhubframework: In here it does not make sense. This is service and add/remove operation on list don't change data in database. Always this is a new instance of list.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
openhubframework: For me it is important have possibility change name and code. When i create node with bad code or name (typing error) my only posibility would be delete node and create new one. This is not user friendly.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
openhubframework: You mean substring? Trim not help with too long string.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
pjuza: I agree with this possibility from user's point of view but even then object can be immutable ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
pjuza: Yes, substring ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
pjuza: You can use toString() in Enum if you have more attributes ...

core/src/main/java/org/openhubframework/openhub/core/common/asynch/queue/PartlyFailedMessagesPoolRoute.java
```diff
@@ -72,22 +72,27 @@ public final void configure() throws Exception {
...
-     * Checks if ESB goes down or not.
+     * Checks if actual node process existing message.
      *
-     * @return {@code true} if ESB is in "stopping mode" otherwise {@code false}
+     * @return {@code true} is process existing message, {@code false} - otherwise
      */
     @Handler
-    public boolean isNotInStoppingMode() {
-        StopService stopService = getApplicationContext().getBean(StopService.class);
+    public boolean isProcessExistingMessage() {
```
pjuza: Better name should be isProcessingExistingMessage

web-admin/src/main/java/org/openhubframework/openhub/admin/web/StopController.java
```diff
@@ -69,19 +74,36 @@ public String stopEsb(ModelMap model) {
...
         model.addAttribute("processingCount", messageService.getCountMessages(MsgStateEnum.PROCESSING, null));
         model.addAttribute("waitingForResCount", messageService.getCountMessages(MsgStateEnum.WAITING_FOR_RES, null));
     }
+
+    /**
+     * Gets if all nodes are in {@link NodeState#STOPPED}.
+     *
+     * @return {@code true} all node is in {@link NodeState#STOPPED}, {@code false} - otherwise
+     */
+    private boolean isNodesInStopping() {
```
pjuza: Wouldn't be isNodeInStopping() better name?

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
pjuza: I agree ...

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Implementation of the {@link NodeService} interface for manipulating with {@link Node}.
+ *
+ * @author Roman Havlicek
+ * @see Node
+ * @since 2.0.0
+ */
+@Service
+public class NodeServiceImpl implements NodeService, ConfCheck {
```
pjuza: I would prefer to have all this implementation directly in Node object but I agree that we should add AOP support (https://openhubframework.atlassian.net/browse/OHFJIRA-32).

core/src/main/java/org/openhubframework/openhub/core/common/dao/NodeDao.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    List<Node> getAllNodes();
+
+    /**
+     * Gets node by identifier.
+     *
+     * @param nodeId node identifier
+     * @return found node
+     */
+    Node getNodeById(Long nodeId);
```
pjuza, line 53: Do you really need this method?
I see you have nodeId (long) in Message object => wouldn't be better to change it to object approach = Message.getNode istead of Message.getNodeId ?

pjuza (commented):
I see lot of good changes in the right way but there are still several points from review for discussion.
Please go through all comments and write me your feedback.

core/src/main/java/org/openhubframework/openhub/core/common/dao/NodeDao.java
openhubframework: This method is prepared for view/edit in web client (like delete method, or getAllNodes() - we don't need them yet).
In message i will add node property with set/get methods.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/StopController.java
openhubframework: It would be better. I will rename it.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/queue/PartlyFailedMessagesPoolRoute.java
openhubframework: It would be better. I will rename it.

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
openhubframework: After we will have AOP support in project i will move these method into Node class.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/StopController.java
openhubframework: This method checks all nodes state and return if all nodes is in stopping. I will rename it to isAllNodesInStopping - this would be better

openhubframework:
Changes are ready to review.

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
```diff
@@ -142,6 +142,10 @@
      */
     public static final String ALERTS_REPEAT_TIME_SEC = PREFIX + "alerts.repeatTimeSec";
 
+    /**
+     * Code of actual node for this application server instance.
+     */
+    public static final String NODE_ACTUAL_INSTANCE_CODE = PREFIX + "node.actualInstance.code";
```
hanusto: In logging improvement we added `ohf.cluster.nodeid` property as identifier of node in clustered environment. We should consolidate that. This property is used in `ohf-logback.xml` file.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
```diff
@@ -762,6 +768,36 @@ public void setProcessingPriority(int processingPriority) {
...
+        return nodeId;
+    }
+
+    /**
+     * Gets node that process this message.
+     *
+     * @return node, {@code NULL} no node process this message
+     */
+    @Nullable
+    public Node getNode() {
```
hanusto: Why `getNode()` *and also* `getNodeId()`? It is similar getter (node.id), i prefer first one.

pjuza:
It looks fine but I have last three questions please:
- isn't ClusterNode better than Node? :)
- why is varchar(64) for code in node table? I remember that it's better to have small fixed character size for types which would be used for often searching (it's better for indexing). Maybe I'm wrong ...
- wouldn't be better to have immutable Node object? Instead of setCode() method create new object ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
hanusto: This is not change of this PR, but I think no. We should use only PK or natural mandatory identifiers.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
hanusto: I agree with Petr.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * @author Roman Havlicek
+ * @since 2.0.0
+ */
+@Entity
+@Table(name = "node",
+        uniqueConstraints = {
+                @UniqueConstraint(name = "uq_node_code", columnNames = {"code"}),
+                @UniqueConstraint(name = "uq_node_name", columnNames = {"name"})})
+public class Node implements HumanReadable {
```
hanusto: I suggest to rename all status-based methods, for example `setRunState`. From `run` (or `process`) method I expect sophisticated behaviour than only change of state.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
hanusto: I imagine some codes (and names) but I think that in most cases it should be node1, node2 or something similar. I don`t think that this attribute is important from user point of view...

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+ * Contains all states for {@link Node}
+ *
+ * @author Roman Havlicek
+ * @see Node#state
+ * @since 2.0.0
+ */
+public enum NodeState {
+
+    /**
+     * Node process new and existing (saved messages in ESB) messages.
```
hanusto: Cluster node `process`, or better verb is `handles`.

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    RUN(true, true),
+
+    /**
+     * Node process only existing messages. New messages/requests are rejected.
```
hanusto: handles...

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    PROCESS_EXISTING_MESSAGES(false, true),
+
+    /**
+     * Node not process existing messages and new messages.
```
hanusto: handles...

core-api/src/main/java/org/openhubframework/openhub/api/entity/Node.java
hanusto: I also recommend to rename `process` to `handle`.

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * {@code true} node process existing saved messages, {@code false} - otherwise.
+     */
+    private final boolean processExistingMessages;
+
+    /**
+     * New instance.
+     *
+     * @param processNewMessages      @code true} node process new incoming messages, {@code false} - otherwise
```
hanusto: probably lost `{`

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public boolean isProcessExistingMessages() {
+        return processExistingMessages;
+    }
+
+    //--------------------------------------------- TOSTRING / HASH / EQUALS -------------------------------------------
+
+    @Override
+    public String toString() {
+        return new ToStringBuilder(this)
```
hanusto: `return new ToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE)`

core-spi/src/main/java/org/openhubframework/openhub/spi/node/NodeService.java
```diff
@@ -0,0 +1,58 @@
...
+ * @since 2.0.0
+ */
+public interface NodeService {
+
+    /**
+     * Insert new {@link Node}.
+     *
+     * @param node node
+     */
+    void insertNode(Node node);
```
hanusto: We can use only `insert` method, without node, it is domain of service. This method should returns inserted `Node` entity for example where node id is generated. It is the same of update method. It is good approach and design pattern to return updated object.

core-api/src/main/java/org/openhubframework/openhub/api/entity/NodeState.java
```diff
@@ -0,0 +1,78 @@
...
+        this.processNewMessages = processNewMessages;
+        this.processExistingMessages = processExistingMessages;
+    }
+
+    /**
+     * Gets if {@link Node} in this state process new incoming messages.
+     *
+     * @return {@code true} node process new incoming messages, {@code false} - oterwise
+     */
+    public boolean isProcessNewMessages() {
```
hanusto: for all getters = Gets `true` if . And maybe better name is: `isAbleToProcessNewMessages` or something similar. The same for method below...

core/src/main/java/org/openhubframework/openhub/core/common/asynch/AsynchMessageRoute.java
```diff
@@ -478,4 +491,18 @@ public void changeStateMessageToInQueue(@Body Message msg) throws LockFailureExc
                     + "': " + msg.toHumanString());
         }
     }
+
+    /**
+     * Checks if actual node process existing message.
+     *
+     * @throws StoppingException if node not proess existing message
```
hanusto: typo `proess`, therefore I recommend to use `handle` verbs :)

core/src/main/java/org/openhubframework/openhub/core/common/dao/NodeDaoJpaImpl.java
```diff
@@ -0,0 +1,95 @@
...
+ * @since 2.0.0
+ */
+@Repository
+public class NodeDaoJpaImpl implements NodeDao {
+
+    @PersistenceContext(unitName = DbConst.UNIT_NAME)
+    private EntityManager em;
+
+    @Override
+    @Transactional(propagation = Propagation.MANDATORY)
```
hanusto, line 35: Is this necessary? Service layer is marked as transaction, so it is ok. Additional mandatory propagation decreases performance.

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(NodeServiceImpl.class);
+
+    private final TransactionTemplate transactionTemplate;
+
+    @Autowired
+    private NodeDao nodeDao;
+
+    @Value("${" + CoreProps.NODE_ACTUAL_INSTANCE_CODE + "}")
+    private String actualNodeCode;
```
hanusto: Why `ConfigurationItem` approach is not used?

```
@ConfigurableValue(key = CoreProps.NODE_ACTUAL_INSTANCE_CODE)
private ConfigurationItem<String> actualNodeCode;
```

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Contains actual node for this instance application server.
```
hanusto: I would remove `application server`. OpenHub can runs as simple microservice without application server.

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+import org.openhubframework.openhub.spi.node.NodeService;
+
+/**
+ * Implementation of the {@link NodeService} interface for manipulating with {@link Node}.
+ *
+ * @author Roman Havlicek
+ * @see Node
+ * @since 2.0.0
+ */
+@Service
```
hanusto: `@Transactional(readOnly = true)`

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    @Transactional
+    public void updateNode(Node node) {
+        Assert.notNull(node, "node must not be null");
+        Assert.notNull(node.getNodeId(), "only new node can be update");
+
+        nodeDao.update(node);
+
+        if (node.getNodeId().equals(getActualNode().getNodeId())) {
```
hanusto: Why node#equals is not used?

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
hanusto: But **unmodified** design is good approach for caller. From caller point of view we dont know implementation detail.

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    public Node getActualNode() {
+        if (actualNode == null) {
+            check();
+        }
+        return actualNode;
+    }
+
+    @Override
+    public synchronized void check() throws ConfigurationException {
+        //test if configuration value exist
```
hanusto, line 125: We should perform double check: if `getActualNode()` returns node (instead of `null`), we can skip next processing. This section is synchronized, it is ok, but we can have situation when two one thread is suspended and waits for unlock previously acquired thread.

hanusto (changes requested):
I did review. Concept of solution for node evidence is ok. We should discuss or changed something according to proposal. After that everything needed should be documented. The most important is PR description with link reference to JIRA issue (see for example https://github.com/OpenWiseSolutions/openhub-framework/pull/27). There are also many recommendation for renaming, for example `process` to `handle`. Thanks for your clarification.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
openhubframework: There is getNodeId because method getNode() gets data from database. When someone needs only node identifier then call method getNodeId - without load data from database.

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
openhubframework: This control in getAcutalNode is only for tests, because in tests method check() is not called. In production environment actualNode attribute never be null.

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
openhubframework: I will set in ohf-logback.xml node property on ohf.node.actualInstance.code

core/src/main/java/org/openhubframework/openhub/core/common/dao/NodeDaoJpaImpl.java
openhubframework: Is really performance decreases so big? I am calmer if in dao insert/update/delete methods is transaction exist control.

hanusto: For big load yes. There must be additional extra check for transaction...

core/src/main/java/org/openhubframework/openhub/core/node/NodeServiceImpl.java
hanusto: Sorry I don't understand that

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
```diff
@@ -813,6 +849,7 @@ public String toString() {
             .append("guaranteedOrder", guaranteedOrder)
             .append("excludeFailedState", excludeFailedState)
             .append("processingPriority", processingPriority)
+            .append("nodeId", nodeId)
```
hanusto, line 835: getNodeId() is safer because there is many-to-one relation with lazy type so node is proxied -> we should ever call methods instead fields.

core-api/src/main/java/org/openhubframework/openhub/api/entity/Message.java
hanusto: Ok, I understand this optimalization but you should documented it why this approach is used.

core-spi/src/main/java/org/openhubframework/openhub/spi/node/ChangeNode.java
```diff
@@ -0,0 +1,24 @@
...
+ * Callback function for update {@link Node} in {@link NodeService#update(Node, ChangeNode)}.
+ * In method {@link #updateNode(MutableNode)} we can change all node attributes.
+ *
+ * @author Roman Havlicek
+ * @see NodeService#update(Node, ChangeNode)
+ * @see Node
+ * @see MutableNode
+ * @since 2.0
+ */
+public interface ChangeNode {
```
pjuza: What about ChangeNodeCallback or ChangeNodeSetter (inspiration by JdbcOperations from Spring)

core/src/main/java/org/openhubframework/openhub/core/common/dao/NodeDao.java
```diff
@@ -0,0 +1,69 @@
...
+ * @since 2.0
+ */
+public interface NodeDao {
+
+    /**
+     * Insert new {@link MutableNode}.
+     *
+     * @param node node
+     */
+    void insert(MutableNode node);
```
pjuza: Why don't use Node interface only?

core/src/main/java/org/openhubframework/openhub/core/node/ImmutableNode.java
```diff
@@ -0,0 +1,190 @@
...
+        this.code = code;
+        this.name = name;
+        this.description = description;
+        this.state = state;
+    }
+
+    //-------------------------------------------------- STATE ---------------------------------------------------------
+
+    @Override
+    public boolean isStopped() {
```
pjuza: I would create AbstractNode with common functions (e.g. state functions) for ImmutableNode and MutableNode ...
Maybe you could add all attributes (= JPA mappings) to this abstract class and use this abstract class in DAO layer.

core-api/src/main/java/org/openhubframework/openhub/api/entity/MutableNode.java
```diff
@@ -0,0 +1,247 @@
...
+        this.name = name;
+        this.state = state;
+    }
+
+    //-------------------------------------------------- STATE ---------------------------------------------------------
+
+    /**
+     * Node handles all messages (existing and new) {@link NodeState#RUN}.
+     */
+    public void setRunState() {
```
pjuza, line 98: Why these state() methods aren't in interface Node (or create new interface NodeLifecycle with these state methods only). 
I miss validations if it's possible to set specific state when node is in another state.
If there is flow of node's state then I would better describe it.

pjuza (changes requested):
I like these changes with immutable vs mutable node. 
I would think about if common abstract node class isn't cleaner solution, also with using in DAO layer.

core-api/src/main/java/org/openhubframework/openhub/api/entity/MutableNode.java
openhubframework: Because Node interface is only for reading. Node interface can not have setter, bacause it is immutable.
Validations it is not needed. There is no flow, you can change node state from any into any state.

core/src/main/java/org/openhubframework/openhub/core/node/ImmutableNode.java
openhubframework: AbstractNode can not have attributes, because in immutable node is all attributes final. AbstractNode is good idea, i will create it.

core-spi/src/main/java/org/openhubframework/openhub/spi/node/ChangeNode.java
openhubframework: You have right ChangeNodeCallback will be better. I will rename it

openhubframework:
It is prepare for review:

- I created new class AbstractNode
- In NodeDao interface is now work only with Node interface

pjuza:
You wrote "AbstractNode can not have attributes, because in immutable node is all attributes final." -  I don't think so because you don't have to have final modifier, you can use suitable constructor and no setter methods.

pjuza (approved):
Good job, I approve it ...

openhubframework:
"I don't think so because you don't have to have final modifier, you can use suitable constructor and no setter methods." - This is true, but if you dont have final modifier, you can change attribute values with reflection.

hanusto (approved):

hanusto:
We will have to update documentation. For now, I merged that.

PR #33, opened 2017-03-12
OHFJIRA-27 - Add support for clustered jobs (quartz)

openhubframework:
In cluster we need two types of job:
-  jobs which can run on random node and can be executed parallel (it's not condition to start jobs in the same time)
- jobs which can run only once in the cluster, doesn't matter on which node 

Solution:
- remove apache camel quartz component
- jobs and triggers is defined with annotation QuartzJob (use only on method that will be invoked), QuartzSimpleTrigger and QuartzCronTrigger
- in annotation QuartzJob we cen set enum JobExecuteTypeInCluster that has two options:
  * CONCURRENT - job will be executed parralel on nodes
  * NOT_CONCURRENT - job which can run only on one node
- information about concurrent jobs is stored in memory
- information about not concurrent jobs is stored in database

JIRA:
https://openhubframework.atlassian.net/browse/OHFJIRA-27


core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/ClusterExecuteType.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Contains all possibility of running jobs in cluster mode.
+ *
+ * @author Roman Havlicek
+ * @see QuartzJob
+ * @see QuartzCronTrigger
+ * @see QuartzSimpleTrigger
+ * @since 2.0
+ */
+public enum ClusterExecuteType {
```
pjuza: Wouldn't be better ScheduledJobType?

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzSimpleTrigger.java
```diff
@@ -0,0 +1,75 @@
+package org.openhubframework.openhub.api.common.quartz;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import org.apache.commons.lang3.StringUtils;
+import org.quartz.SimpleTrigger;
+
+/**
+ * Defined one {@link SimpleTrigger} in {@link QuartzJob}.
```
pjuza, line 10: Better describe why use this annotation instead of QuartzCronTrigger

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/SimpleTriggerMisfireInstruction.java
```diff
@@ -0,0 +1,91 @@
...
+ * @see QuartzCronTrigger
+ * @see QuartzSimpleTrigger
+ * @since 2.0
+ */
+public enum SimpleTriggerMisfireInstruction {
+
+    /**
+     * See documentation in {@link SimpleTrigger#MISFIRE_INSTRUCTION_SMART_POLICY}.
+     *
+     * @see SimpleTrigger
```
pjuza: Isn't it needless to add more times "@see SimpleTrigger"?

core-spi/src/main/java/org/openhubframework/openhub/spi/quartz/TriggerFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.openhubframework.openhub.api.common.quartz.ClusterExecuteType;
+import org.openhubframework.openhub.api.common.quartz.QuartzCronTrigger;
+import org.openhubframework.openhub.api.common.quartz.QuartzJob;
+import org.openhubframework.openhub.api.common.quartz.QuartzSimpleTrigger;
+
+/**
+ * Factory for create new {@link Trigger}s instance.
+ * <p>
+ * Every trigger instance running concurrent in all nodes in cluster at the same time
```
pjuza: language mistake - Every trigger instance is able to running in ...

Factory for create new => Factory class for creating new ...

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzSimpleTrigger.java
```diff
@@ -0,0 +1,75 @@
...
+     * @return misfire instruction
+     */
+    SimpleTriggerMisfireInstruction misfireInstruction() default SimpleTriggerMisfireInstruction.SMART_POLICY;
+
+    /**
+     * Gets execute type in cluster mode.
+     *
+     * @return execute type in cluster mode
+     */
+    ClusterExecuteType clusterExecuteType();
```
pjuza: Why is there not default value? I would expect concurrent ...

core/src/main/java/org/openhubframework/openhub/core/common/asynch/queue/PartlyFailedMessagesJob.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Route definition that starts job process that pools message queue (=database)
+ * and takes {@link MsgStateEnum#PARTLY_FAILED} messages for further processing.
+ * <p>
+ * Repeat interval for this job is load from configuration {@value CoreProps#ASYNCH_PARTLY_FAILED_REPEAT_TIME_SEC}.
+ * Job running concurrent in all nodes in cluster at the same time.
+ * </p>
+ *
+ * @author Petr Juza
```
pjuza, line 43: I don't know why there are so many @see items in specific implementations, I would remove them.

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return {@code true} uses {@link DisallowConcurrentExecution}, {@code false} - otherwise
+     */
+    boolean stateful() default false;
+
+    /**
+     * All cron triggers definition for this job.
+     *
+     * @return cron triggers
+     */
+    QuartzCronTrigger[] cronTriggers() default {};
```
pjuza, line 90: Is it possible to set cron and simple triggers at the same time?
If not then add description to JavaDoc + it's possible to define common parent class/annotation and then have only one trigger parameter?

core/src/main/java/org/openhubframework/openhub/core/common/quartz/OpenHubQuartzJob.java
```diff
@@ -0,0 +1,73 @@
...
+import org.quartz.DisallowConcurrentExecution;
+import org.quartz.Job;
+
+import org.openhubframework.openhub.api.common.quartz.QuartzCronTrigger;
+import org.openhubframework.openhub.api.common.quartz.QuartzJob;
+import org.openhubframework.openhub.api.common.quartz.QuartzSimpleTrigger;
+
+/**
+ * Defined one quartz {@link Job} with default group name {@value #OPEN_HUB_JOB_GROUP_NAME}.
+ * For base use see in {@link {@link QuartzJob}}.
```
pjuza: Remove once @link ...

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
```diff
@@ -0,0 +1,87 @@
...
+     * @return group name
+     */
+    String group() default StringUtils.EMPTY;
+
+    /**
+     * Job name.
+     *
+     * @return name
+     */
+    String name() default StringUtils.EMPTY;
```
pjuza: I see method in QuartzUtils for generating job's name. Wouldn't be better to define JobNameGenerator interface with default implementation from QuartzUtils?

core/src/main/java/org/openhubframework/openhub/core/config/QuartzConfig.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Properties for cluster scheduler for Postgres database.
+     *
+     * @return properties for Postgres database
+     */
+    @Bean(name = "quartzProperties")
+    @Profile(Profiles.POSTGRES)
+    public Properties quartzPostgresProperties() {
```
pjuza: I would define it in property file because when you add new DB then you should always add new method to the code.

pjuza (changes requested):
Quite big implementation :).

I don't have any critical comments, only many smallers. I'm not able to validate the whole model (because it's too big ...) but I believe it's ok ...

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
openhubframework: Yes it is. One job can have one or more QuartzCronTrigger and one or more QuartzSimpleTrigger at the same time.

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
openhubframework: In QuartzUtils is not method for generate job name, but for generate trigger name. Job name is defined in annotation attribute name. I will create TriggerNameGenerator interface.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/queue/PartlyFailedMessagesJob.java
openhubframework: I will remove it

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/SimpleTriggerMisfireInstruction.java
openhubframework: I will remove it and add "@see SimpleTrigger" on class.

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/ClusterExecuteType.java
openhubframework: I dont know. This is not for job, but it is for trigger. And maybe this is not good. I will add this setting into job. And name can be JobExecuteTypeInCluster.

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzSimpleTrigger.java
openhubframework: I think that default value is not good idea. Set this attribute is important and when he has default value, many user not fill them.

pjuza (commented):
I went through all your comments and it's OK for me.

Only one thing - how to solve when new database type will be used? At this moment you should add new code and it's not flexible ...

openhubframework:
You have right, this is not good solution. I will change it and move properties into configuration file application-h2.cfg or application-postgres.cfg.

openhubframework:
It is ready for review.

Changes:

- Move quartz settgins from QuartzConfig into application properties

- Add name() and group() into QuartzSimpleTrigger and QuartzCronTrigger - now user can direct specify trigger name and trigger group (generator is not necessary)
- Edit javadoc (remove @see etc.)
- rename ClusterExecuteType into JobExecuteTypeInCluster and add this attribute into QuartzJob (remove from triggers)

common/src/main/java/org/openhubframework/openhub/common/Tools.java
```diff
@@ -210,6 +211,31 @@ public static OffsetDateTime fromUTC(long utcMillis) {
...
 
     /**
+     * Gets {@link Properties} from {@link ConfigurableEnvironment} with properties that name start with
+     * prefix from parameter.
+     *
+     * @param env                environment
+     * @param propertyNamePrefix prefix for property names
+     * @return {@link Properties} with properties that name start with propertyNamePrefix
+     */
+    public static Properties getPropertiesWithPrefix(ConfigurableEnvironment env, String propertyNamePrefix) {
```
pjuza, line 221: Consider to use Tools.getAllKnownPropertyNames in implementation ...

pjuza (commented):
Two common points:
- test coverage descreased about 6% - it's too much, please add more unit tests
- please add wiki pages with information about using your quartz API. It will be helpful to better understand your code

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.quartz.DisallowConcurrentExecution;
+import org.quartz.Job;
+
+/**
+ * Defined one quartz {@link Job} with simple trigger ({@link QuartzSimpleTrigger}) or cron trigger
+ * {@link QuartzCronTrigger}.
+ * <p>
+ * Annotation can be added on method and this method will be called when trigger will be executed.
+ * </p>
```
pjuza, line 14: if you have starting and ending element p then it will create two empty lines and you propably want to have one line, don't you?

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>
+ * Every job must have at least one {@link QuartzCronTrigger} or {@link QuartzSimpleTrigger}. And one job can have
```
pjuza, line 16: it's quite complicated sentence :)

core/src/main/java/org/openhubframework/openhub/core/common/quartz/jobfactory/BeanMethodJobFactory.java
```diff
@@ -0,0 +1,75 @@
...
+                ((MethodInvokingJobDetailFactoryBean.MethodInvokingJob) job).setMethodInvoker(methodInvoker);
+            }
+        }
+        return job;
+    }
+
+    //-------------------------------------------------- SET / GET -----------------------------------------------------
+
+    @Override
+    public void setApplicationContext(final ApplicationContext applicationContext) {
```
pjuza: We have already full AOP support thus you can use @Configured ...

pjuza (commented):
Once again review ...

common/src/main/java/org/openhubframework/openhub/common/Tools.java
openhubframework: I will use it. One small disadvantage is, that list will be use twice (first for getting names and second for getting values for names).

core-api/src/main/java/org/openhubframework/openhub/api/common/quartz/QuartzJob.java
openhubframework: I dont understand. Where is two empty lines? In HTML or in source code?

pjuza: When you use ending p and starting p ...

core/src/main/java/org/openhubframework/openhub/core/common/quartz/jobfactory/BeanMethodJobFactory.java
openhubframework: I used Autowired for load ApplicationContext property

hanusto (approved):
For this moment these PR meets the requirements.

PR #34, opened 2017-03-13
Add data into cache and optimalize loading properties from DB

pjuza:
This task is mainly about using cache for caching configuration parameters. 
By default I set cache "config_params" as read-only with timeouts for eviction (see ohf_hazelcast.xml). 

There are several possibilities how to monitor cache state:
- use management center of Hazelcast (http://docs.hazelcast.org/docs/latest/manual/html-single/index.html#managing-maps)
- there is JMX API of Hazelcast (http://docs.hazelcast.org/docs/latest/manual/html-single/index.html#jmx-api-per-member)
- there are cache metrics in Spring Boot Actuator (https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-datasource-cache)

It would be useful to add possibility to evict/clear cache values into admin GUI (https://openhubframework.atlassian.net/browse/OHFJIRA-58).

Shortly:
- added caching for configuration parameters
- improved condition if req/res should be saved
- corrected initialization of ConfigurationChecker
- added new property ohf.dbProperty.includePattern that defines pattern for filtering which property names should be loaded from DB

Issue: OHFJIRA-37


core/src/main/java/org/openhubframework/openhub/core/configuration/AutoConfigurationItemProperties.java
```diff
@@ -65,6 +68,7 @@ public ConfigurationService configurationService(final Environment environment)
         return new ConfigurationService() {
 
             @Override
+            @Cacheable(value = CacheNames.CONFIG_PARAMS)
```
hanusto: better will be `@Cacheable(CacheNames.CONFIG_PARAMS)` or `@Cacheable(cacheNames = CacheNames.CONFIG_PARAMS)`

core/src/main/java/org/openhubframework/openhub/core/configuration/DbExternalPropertiesAutoConfiguration.java
```diff
@@ -52,6 +61,9 @@
 
     public static final String DB_CONF_PROPERTY_SOURCE_NAME = "dbConfiguration";
 
+    @Value("${" + PROPERTY_INCLUDE_PATTERN + ":#{null}}")
```
hanusto, line 64: Why not `ConfigurationItem`?

hanusto (approved):
Great, only some recommendations.

core/src/main/java/org/openhubframework/openhub/core/configuration/AutoConfigurationItemProperties.java
pjuza: OK, I will change it ...

core/src/main/java/org/openhubframework/openhub/core/configuration/DbExternalPropertiesAutoConfiguration.java
pjuza: I didn't have idea that I could use ConfigurationItem when it's implementation for loading configuration parameters that ConfigurationItem uses :).
Does it make sense here? Maybe code would be pretty but nothing else ...

hanusto: yes, sorry, my mistake

pjuza:
I pushed @Cacheable correction. Please merge this PR.

hanusto:
Done

PR #35, opened 2017-03-14
OHFJIRA-54: Environment and config properties overview

hanusto:
From admin GUI point of view we have to properly configure javamelody stack. Actually javamelody monitoring is disabled but it can be enabled via `javamelody.enabled = true` property. Javamelody statistics are served by default at /web/admin/monitoring/javamelody.

Endpoints to show actually used config props are exposed by Spring Boot.

[Documentation](https://openhubframework.atlassian.net/wiki/display/OHFV1/JavaMelody) is updated.

Issue: [OHFJIRA-54](https://openhubframework.atlassian.net/browse/OHFJIRA-54)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,7 +150,7 @@
         <postgresql-version>9.4-1201-jdbc41</postgresql-version>
         <oracle-version>12.1.0.2</oracle-version>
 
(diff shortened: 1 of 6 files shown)


pjuza (approved):
It's OK ...

PR #39, opened 2017-04-02
ConfigurationChecker with embedded servlet container

kkovarik:
### ISSUE ADDRESSED
* when using embedded tomcat & default db configuration (ohf.server.localhostUri.check = true), OpenHub won´t start, as ConfigurationChecker tries to ping server, however tomcat is not yet started, resulting in following exception:
`org.openhubframework.openhub.api.exception.ConfigurationException: Configuration error - 
parameter 'ohf.server.localhostUri' with value 'http://localhost:8080' is probably wrong, 
URI isn't reachable.`

### CHANGES
* ConfigurationChecker now listens to ApplicationReadyEvent from Spring, http://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/context/event/ApplicationReadyEvent.html
* related issue on StackOverflow: http://stackoverflow.com/questions/28240403/spring-boot-wait-for-web-server-to-start

### DIFF
--- a/core/src/main/java/org/openhubframework/openhub/core/confcheck/ConfigurationChecker.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/confcheck/ConfigurationChecker.java
@@ -35,10 +35,10 @@ import org.apache.http.impl.client.HttpClients;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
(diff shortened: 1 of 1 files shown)


pjuza (approved):
This is very old functionality when there was no Spring Boot and this event. I agree and thank you for improving it.

hanusto (approved):
We have to merge previous PR which contains necessary changes to FIX wrong tests. Thanks for you contribution.

kkovarik:
Thanks for approves, @hanusto sure I will rebase it when its merged...

hanusto:
I did it :).

PR #40, opened 2017-04-02
OHFJIRA-10 Flyway db-migration

kkovarik:
### INFO
* openhub db schema migration handled by Flyway

### CHANGES
* flyway enabled:
    * added sql scripts for H2 & PostgreSQL, all current db tests are passing
    * spring JDBC initialization disabled
    * hibernate switched to validation mode
    * minor update in db_schema_postgreSql_archive script
    * ~~fixed PartlyFailedMessagesPoolRouteTest (was not using h2 profile)~~
* added new sequence for Node table
* ~~added sample docker image for postgresql (as I used for testing of db migration)~~

### RELATED JIRA
* https://openhubframework.atlassian.net/browse/OHFJIRA-10

### OPEN QUESTIONS
* H2 was generated from entities by hibernate, now disabled and switched to Flyway to keep things consist, as side effect two SQL schemas (postgres & h2) needs to be maintaned, is it ok? 
* what with existing databases?
* implementations of OpenHub, will there be some mechanism for their custom scripts? at least simple conventions like that OpenHub will use only defined versions, so implementations can add their own after them?
* db migration tests against PostgreSQL database - using docker on Travis?

## Using
Before first startup of OpenHub after this PR run OpenHub with `flyway.baseline-on-migrate=true`. Thanks to that new table schema_version table will be created and all existed scrips resolved as applied. New added script will have **V1_0_3** version name.


hanusto:
@kkovarik 
OP: 
ad1) I think that H2 script is preferred, because schema is isolated
ad2) OpenHub among 1.0 and 2.0 version has big incompatible changes, so from my point of view we can do agreement, support of existing databases is out of scope. Version 1.0 is not in production, so it is ok.
ad3) in future yes, but now? I dont know. I agree with your idea.
ad4) if you know how to, yes, because we have to sure, that all DB scripts are valid. But it is "bonus".

hanusto:
Maybe can you please compare two approaches: flyway and liquibase? Liquibase supports multiple database: http://www.liquibase.org/databases.html with one code.

kkovarik:
@hanusto @pjuza @openhubframework  Please look at the task comments in JIRA https://openhubframework.atlassian.net/browse/OHFJIRA-10 to discuss which Flyway or Liquibase approach to be used...

hanusto:
I saw your analysis in JIRA issue. I am for Flyway.

kkovarik:
Thanks for the feedback, PR updated, ready for review.

* flyway is enabled, all scripts moved and consolidated in two files for each db engine. spring jpa datasource initialization disabled.
* tested both on H2 and PostgreSQL engines, all DB tests are passing, flyway is used here too.
* one change in schema: the NODE table was using default hibernate_sequence, changed to custom sequence (openhub_node_sequence). Did not wanted to reuse the openhub_sequence for this technical ID, to be shared with messages etc.

pjuza:
Answer to question 3 - custom sql script: everyone have possibility to change application*.properties files and can set/add own sql scripts:

`a schema (DDL) script resource reference`
`spring.datasource.schema=classpath:db/db_schema_h2.sql`
`a data (DML) script resource reference`
`spring.datasource.data=classpath:/db/db_init-configuration.sql`

core-api/src/main/java/org/openhubframework/openhub/api/entity/MutableNode.java
```diff
@@ -29,7 +29,8 @@
      */
     @Id
     @Column(name = "node_id")
-    @GeneratedValue(strategy = GenerationType.AUTO)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "openhub_node_id_sequence")
```
pjuza, line 32: It's question for Radek why common hibernate sequence was used ...

test/src/main/resources/application-test.properties
```diff
@@ -28,13 +28,14 @@ logging.level.org.apache.camel=ERROR
 # = JPA / HIBERNATE
 # ===============================
 
-# Initialize the schema on startup.
-spring.jpa.generate-ddl=true
+# Initialize the schema with jpa is disabled.
+spring.jpa.generate-ddl=false
```
pjuza, line 32: I'm not sure about that - why we won't use DDL auto-generation for H2 - even for production code? Now we have to maintain one more sql script and we will lost flexibility of H2 auto-generation. 

Please what is the main reason for that?

pjuza (approved):
I say yes after discussion ...

core-api/src/main/java/org/openhubframework/openhub/api/entity/MutableNode.java
hanusto: @openhubframework can you please explain that? It is mistake or require think?

test/src/main/resources/application-test.properties
hanusto: Reason of discussion: preferred is to use migration capability over auto-generation of DB schema because it is better for maintaining.

hanusto:
My answer for @pjuza related Q3 - probably @kkovarik was talking about possibilities how project can use and add custom flyway scripts. Because there is question - between v2.0 and v3.0 we will add two scripts and project which is using v2.0 wants to add one. In which sequence these script should be applied when project will migrate from v2.0 to v3.0? Probably our first so we have to "recommended steps" how to create custom scripts to avoid errors during migration.

core/pom.xml
```diff
@@ -81,6 +81,11 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-jdbc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.flywaydb</groupId>
+            <artifactId>flyway-core</artifactId>
+            <version>${flyway.version}</version>
```
hanusto: Flyway is a part of spring boot dependency BOM, so you can omit that I thing.

core/src/main/resources/db/migration/h2/V1_1__config_data_init.sql
```diff
@@ -0,0 +1,128 @@
+--
```
hanusto: Is there possibility to rename this file? I am talking about relation between version number of script and version of application. Maybe it is not mandatory, but is config data next version? I am looking at https://flywaydb.org/documentation/migration/versioned and there is example and [lessons learned](http://www.jeremyjarrell.com/using-flyway-db-with-distributed-version-control/) -> **Prefix your migrations with timestamps rather than integers**.

Actually version 1 is released so we can use prefix _V1_0_0_ for all scripts. Third number will be used for DB script versions or better say "scripts". What do you thing?

core/src/main/resources/db/migration/postgresql/V1_0__schema_init.sql
```diff
@@ -1,11 +1,183 @@
 --
--- DB increment script for version 0.4 - archive functionality that is compliant with 0.4 DB schema
```
hanusto: :1st_place_medal:

core/src/main/resources/db/readme.txt
```diff
@@ -1,3 +1,7 @@
-Initialization SQL scripts for specific databases.
```
hanusto, line 1: I prefer to use `readme.md` name with markdown syntax.

web-admin/src/main/resources/application.properties
```diff
@@ -300,6 +300,34 @@ org.quartz.jobStore.class = org.quartz.impl.jdbcjobstore.JobStoreTX
 org.quartz.jobStore.useProperties = true
 
 # ===============================
+# = FLYWAY
+# ===============================
+flyway.baseline-description=OpenHub2.0
+#flyway.baseline-version=1 # version to start migration
+#flyway.baseline-on-migrate= #
```
hanusto: This property is very useful when `flyway.baseline-on-migrate=true` if you already have some tables in the database. So we should documentation this property for example in `readme.md` file.

hanusto (approved):
Thank you for your PR. I agree with this improvement, only some comments added - please check that.

One question is - this feature should be documented - is prepared or planned?

core/src/main/resources/db/readme.txt
kkovarik: Good idea, renamed & changed to markdown.

web-admin/src/main/resources/application.properties
kkovarik: Added description to readme.md

core/pom.xml
kkovarik: Thanks for pointing out, removed 👍

core/src/main/resources/db/migration/h2/V1_1__config_data_init.sql
kkovarik: Yes, good versioning convention is important, thanks for pointing it out, I drafted the numbers with intention to revisit it again. Cannot use prefix V1_0_0 for all of them, the numbering must be non-conflicting. So now I am at V1_0_0, V1_0_1, V1_0_2...

I have read the lessons learned article, it made kind of sense, however timestamps seemed like overkill at the time - again it is more like question of how often the scripts will be added so the branch sql-number conflict could occur.
Maybe with dates - instead of timestamps?

web-admin/src/main/resources/application.properties
hanusto: Flyway integration should be documented - is prepared or planned?

kkovarik: Yes, just wondering what it best location - wiki? https://openhubframework.atlassian.net/wiki/display/OHF/Running+OpenHub ?

pjuza: I would suggest to create new page "Flyway" under Data model page (https://openhubframework.atlassian.net/wiki/display/OHF/Data+model) - add some common info about Flyway and then how it's solved in OHF ...

kkovarik: Added some common info, please check if it makes sense, and feel free to suggest anything I missed:
https://openhubframework.atlassian.net/wiki/display/OHF/Flyway

pjuza: It's good, I made few small changes only ...

hanusto:
Please @kkovarik could you do last rebase on develop for merge preparation?

kkovarik:
@hanusto sure thing, rebased on top of current develop

hanusto:
I tested it and I have one proposal. I have fully prepared DB with data, so only one thing what I need is to create `schema_version` table, but actually, when I use `flyway.baseline-on-migrate=true`, base line version is configured (default value) to **1** and all scripts have higher version so all scripts will be applied, but I think it is not required. I think that we should define `flyway.baseline-version=2`. Actual target version is **2** and all scripts are for migration from 1 to 2, but version 1 is not used, so question is if we want to support that.

image

Actually I run OpenHub, startup fails because all scripts are applied (version is higher than baseline). If I change default version (to 2), everything is ok. 

What do you think about that @pjuza @kkovarik ?

kkovarik:
@hanusto well, fine with me, it makes sense, as it is not setup for new project, but there are already developers with own database like yourself. it is easier this way, not to fiddle with version...

hanusto:
Final solution is version 1.0.2 = actual scripts will be resolved as applied and new one will be candidate for migration. Tested.

PR #41, opened 2017-04-03
Minor improvements

kkovarik:
#### CHANGES
* DataTypeConverterTest updated float converter test to respect Locale separator, as it was failing on some developer environment setup
* Git-commit-id-plugin setup to ignore missing .git directory. If openhub project was checked out by using "Download as ZIP" feature of github, project build should not fail just because .git folder is missing


hanusto (approved):
Thx for contribution.

PR #44, opened 2017-04-04
OHFJIRA-64: Expose interface to provide information about logged user.

hanusto:
Admin console has to show detail about logged user. Now only full name of logged user is required.

Issue: [OHFJIRA-64](https://openhubframework.atlassian.net/browse/OHFJIRA-64)


kkovarik (approved):
Seems fine to me

PR #45, opened 2017-04-05
OHFJIRA-53: error code catalog endpoint

hanusto:
OpenHub console should have page where user can see which exception code can be thrown, its description or what is recommended action. It is based upon not approved PR branch, it will be rebased.

API: http://docs.openhubframework.apiary.io/#reference/0/openhub-error-catalog-endpoint/gets-error-catalog-values

Issue: [OHFJIRA-53](https://openhubframework.atlassian.net/browse/OHFJIRA-53)

### DIFF
--- a/web-admin/src/main/java/org/openhubframework/openhub/admin/web/errorscatalog/rest/ErrorsCatalogController.java
+++ b/web-admin/src/main/java/org/openhubframework/openhub/admin/web/errorscatalog/rest/ErrorsCatalogController.java
@@ -1,0 +1,72 @@
+/*
+ *  Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.admin.web.errorscatalog.rest;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.*;
+
+import org.openhubframework.openhub.admin.services.ErrorCatalogService;
+import org.openhubframework.openhub.admin.web.common.AbstractOhfController;
+import org.openhubframework.openhub.admin.web.errorscatalog.rpc.ErrorCodeRpc;
+import org.openhubframework.openhub.admin.web.errorscatalog.rpc.ErrorsCatalogRpc;
+import org.openhubframework.openhub.api.entity.ErrorsCatalog;
+import org.openhubframework.openhub.api.exception.ErrorExtEnum;
+
+
+/**
+ * REST controller to operate over errors catalog.
+ *
+ * @author Tomas Hanus
+ * @since 2.0
+ */
+@RestController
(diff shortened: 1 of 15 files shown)


web-admin/src/main/java/org/openhubframework/openhub/admin/web/auth/rpc/AuthInfoRpc.java
```diff
@@ -0,0 +1,115 @@
...
+
+
+/**
+ * RPC for {@link Principal user} entity.
+ *
+ * @author Tomas Hanus
+ * @since 2.0
+ */
+@XmlRootElement
+public class AuthInfoRpc {
```
pjuza: Isn't possible to use some parent RPC object?
I created some common parent rpc classes but I don't know if they can be used for your case. Maybe some for readonly rpc classes would be nice ...

web-admin/src/main/java/org/openhubframework/openhub/admin/web/errorscatalog/rest/ErrorsCatalogController.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Serves errors catalog.
+     *
+     * @return list of {@link ErrorsCatalogRpc}
+     */
+    @RequestMapping(method = RequestMethod.GET, produces = {"application/xml", "application/json"})
+    @ResponseStatus(HttpStatus.OK)
+    @ResponseBody
+    public List<ErrorsCatalogRpc> errorsCatalog() {
```
pjuza, line 57: Pagination is not necessary? I would let to client decide ...

pjuza (approved):
I found two small points for improvement but I will let them on your consideration.

pjuza:
I like your annotation to the error code ...

web-admin/src/main/java/org/openhubframework/openhub/admin/web/auth/rpc/AuthInfoRpc.java
hanusto: I know about that but for example this object has no PK.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/errorscatalog/rest/ErrorsCatalogController.java
hanusto: I think I dont know. I can add that, but problem is that response contains many error catalog and each error catalog contains many error codes. So what will be in one page? Client can do pagination over error catalogs, but in response only few error catalogs will be returned (per backend system maybe), but with many of error codes. Maybe in client side, we can add dynamic table with scrollbar for error codes in the error catalog.

pjuza: So let's stay with current solution and we will see from the feedback ...

web-admin/src/main/java/org/openhubframework/openhub/admin/web/auth/rpc/AuthInfoRpc.java
pjuza: Ok, let's leave like that. I will make some adjustments around rpc, so I will look at it ...

kkovarik (approved):

web-admin/src/main/java/org/openhubframework/openhub/admin/web/auth/rpc/AuthInfoRpc.java
hanusto: Ok. Maybe some read-only object.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/errorscatalog/rest/ErrorsCatalogController.java
hanusto: Yes. And error catalogs is not "pageable repository". So it would be difficult how to implement pagination. I think that better will be if for example all error codes will be collapsed and only requested catalog will be expanded.

PR #46, opened 2017-04-09
Error in processing async child message

openhubframework:
In situation that some message has child message and this child message has another child message, then head parent message stays always in state WAITING.

Solution:
- recursively call method setStateOk and setStateFailed for parent message
- change state to OK on parent message only if it is in WAITING state 

JIRA:
https://openhubframework.atlassian.net/browse/OHFJIRA-67

### DIFF
--- a/core/src/main/java/org/openhubframework/openhub/core/common/asynch/msg/MessageServiceImpl.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/common/asynch/msg/MessageServiceImpl.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 
+import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/openhubframework/openhub/core/common/dao/MessageDao.java
```diff
@@ -82,6 +82,14 @@
     List<Message> findChildMessages(Message msg);
 
     /**
+     * Finds all child message for parent message.
```
pjuza, line 85: "Finds all child messages of specified parent message."

core/src/main/java/org/openhubframework/openhub/core/common/asynch/msg/MessageSplitterImpl.java
```diff
@@ -103,6 +103,8 @@ public final void splitMessage(@Header(AsynchConstants.MSG_HEADER) Message paren
 
         // mark original message as parent message
         parentMsg.setParentMessage(true);
+        // message has child so on parent set waiting state
```
pjuza, line 106: // message has child => set parent message's state WAITING

core/src/main/java/org/openhubframework/openhub/core/common/asynch/msg/MessageServiceImpl.java
```diff
@@ -279,6 +284,11 @@ private void setParentMsgFailed(Message msg) {
         messageDao.update(parentMsg);
 
         LOG.debug("State of the parent message " + parentMsg.toHumanString() + " was changed to " + MsgStateEnum.FAILED);
+
+        //if message has parent another parent message then set parent message into FAILED
```
pjuza, line 288: // if message has 'hard' parent message then set parent's state to FAILED too

pjuza (approved):
It's hard to find these kinds of errors - thank you for correction.

hanusto (approved):

PR #47, opened 2017-04-09
OHFJIRA-55 - DB config params endpoint improvements

hanusto:
Best practise is to avoid duplication of PK (id, code) between URL PA…TH and payload.

There should be only one place where unique identifier is, for example to check authorization of requester to perform this action. This change is documented (changed) in Apiary.

Issue: [OHFJIRA-55](https://openhubframework.atlassian.net/browse/OHFJIRA-55)


pjuza (approved):
Thank you for your correction, I agree.

PR #53, opened 2017-05-14
OHFJIRA-51: Cluster node management API.

hanusto:
Exposed endpoint for management cluster node entities via REST calls.

API design: http://docs.openhubframework.apiary.io/#reference/0/openhub-cluster-nodes-management
Mockup design: https://openwise.mybalsamiq.com/projects/openhub/Cluster%20nodes

## Changes

- fixed javadoc for `NodeService` API
- added implementation via `ClusterNodeController`
- added test to verify functionality
- added `CollectionWrapper` class to hold RPC items

Issue: OHFJIRA-51

### DIFF
--- a/web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rest/ClusterNodeController.java
+++ b/web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rest/ClusterNodeController.java
@@ -1,0 +1,118 @@
+/*
+ *  Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.admin.web.cluster.node.rest;
+
+import java.util.List;
+import javax.validation.Valid;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.util.Assert;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.*;
+
+import org.openhubframework.openhub.admin.web.cluster.node.rpc.NodeRpc;
+import org.openhubframework.openhub.admin.web.common.AbstractOhfController;
+import org.openhubframework.openhub.admin.web.common.rpc.CollectionWrapper;
+import org.openhubframework.openhub.api.common.Constraints;
+import org.openhubframework.openhub.api.entity.Node;
+import org.openhubframework.openhub.api.exception.NoDataFoundException;
+import org.openhubframework.openhub.api.exception.validation.ValidationException;
+import org.openhubframework.openhub.spi.node.NodeService;
+
+
+/**
+ * REST controller with operations of {@link org.openhubframework.openhub.api.entity.Node} entity.
+ *
+ * @author Tomas Hanus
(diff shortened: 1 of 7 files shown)


web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rpc/NodeRpc.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /**
+     * Validates RPC (mandatory attributes, correct object state etc.).
+     *
+     * @param errors       Binding/validation errors
+     * @param updateEntity {@code null} if new entity is created otherwise entity that is updated
+     * @throws ValidationException when there is error in input data
+     */
+    public final void preSaveValidation(BindingResult errors, @Nullable Node updateEntity) throws ValidationException {
```
pjuza, line 182: Why this method when you can call directly validate() method?

web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rest/ClusterNodeController.java
```diff
@@ -0,0 +1,118 @@
...
+        Assert.notNull(nodeRpc, "nodeRpc can not be null");
+        if (nodeRpc.getId() != null) {
+            Constraints.state(id.equals(nodeRpc.getId()), "IDs must be equal");
+        }
+
+        // get entity to update it
+        Node dbNode = nodeService.getNodeById(id);
+
+        // save entity
+        nodeService.update(dbNode, new NodeRpc.ChangeCallback(errors, nodeRpc));
```
pjuza, line 104: I see that validation concept doesn't go well with concept of ChangeCallback ... It's statement, I would leave it and we will see ...

web-admin/src/main/java/org/openhubframework/openhub/admin/web/common/rpc/CollectionWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+/**
+ * Collection wrapper of items.
+ *
+ * @param <T> - type of item in collection result
+ * @author Tomas Hanus
+ * @since 2.0
+ */
+public class CollectionWrapper<T> {
```
pjuza, line 35: I like it ... 👍

pjuza (approved):

web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rpc/NodeRpc.java
hanusto: Cannot. Validation is protected and not visible from outside.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rest/ClusterNodeController.java
hanusto: Yes, but I think this is exception :).

web-admin/src/main/java/org/openhubframework/openhub/admin/web/cluster/node/rest/ClusterNodeController.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Updates existing {@link NodeRpc}.
+     *
+     * @param id      as identifier of {@link NodeRpc}
+     * @param nodeRpc The {@link NodeRpc}
+     * @throws ValidationException  if input object has wrong values
+     * @throws NoDataFoundException entity not found for update
+     */
+    @RequestMapping(value = "/{id}", method = RequestMethod.PUT, produces = {"application/xml", "application/json"})
```
kkovarik, line 89: Philosophical: Is the **produces** configuration needed with 204 No Content response? As API consumer won´t be expecting any kind of response, he may not expect to send any Accept* headers?

web-admin/src/main/java/org/openhubframework/openhub/admin/web/common/rpc/CollectionWrapper.java
kkovarik: +1

kkovarik (approved):

PR #54, opened 2017-05-18
OHFJIRA-65: Exposed console config endpoint to initialise GUI menu.

hanusto:
This endpoint should be used after authentication of user when menu of GUI is composed. Endpoint provides configuration of menu items, which should be active and so on.

API: http://docs.openhubframework.apiary.io/#reference/0/openhub-console-config

Issue: [OHFJIRA-65](https://openhubframework.atlassian.net/browse/OHFJIRA-65)


pjuza (approved):

hanusto:
E2E tested with GUI - passed.

PR #55, opened 2017-05-18
Fixed problem with serialization REST objects into XML format.

hanusto:
To whole support of XML serialization it is necessary to include [jackson-dataformat-xml](https://github.com/FasterXML/jackson-dataformat-xml) dependency otherwise there is for example problem with Serializable interface (BaseRpc).

See [How to create a SpringBoot XML REST Service](http://javasampleapproach.com/spring-framework/spring-boot/create-springboot-xml-rest-service)

Issue: NOISSUE

### DIFF
--- a/web-admin/pom.xml
+++ b/web-admin/pom.xml
@@ -77,7 +77,11 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
-
+        <!-- XML support-->
+        <dependency>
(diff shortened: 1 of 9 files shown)


kkovarik:
If I understand this correctly, XML serialization with this dependency is switched from jaxb to jackson? 

And the XmlRootElement annotation is not needed anymore? maybe remove it from RPC entities?

hanusto:
Yes, exactly. I removed all JAXB based annotations from RPC objects.

kkovarik (approved):
Seems good to me.

PR #57, opened 2017-05-30
OHFJIRA-52 added wsdl overview

bionaut:


hanusto (approved):
I tested it locally and it works as expected. Thanks.

PR #59, opened 2017-06-18
Catalog API implementation

kkovarik:
### CHANGES
* catalog api implementation added, in minimal scope that is needed for messages (OHFJIRA-57)
* it is fast & simple, only code & description for each catalog entry, no validity, localization or any advanced features.
* see CatalogServiceImplTest for example catalog implementation (SourceSystemTestCatalog)

### API
* http://docs.openhubframework.apiary.io/#reference/0/openhub-catalogs

### DIFF
--- a/core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
+++ b/core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.api.catalog;
+
+import java.util.List;
+
+import org.springframework.aop.framework.AopProxyUtils;
+import org.springframework.core.annotation.AnnotationUtils;
+
+
+/**
+ * Contract for OpenHub catalogs. All catalogs that should be available through
+ * catalog API, should be beans implementing this interface.
+ * Note: designed to be used with helper annotation {@link CatalogComponent}.
+ *
+ * @author Karel Kovarik
+ * @since 2.0
+ * @see CatalogComponent
+ */
+public interface Catalog {
+
+    /**
+     * Get all entries in catalog.
+     */
+    List<CatalogEntry> getEntries();
+
+    /**
+     * Resolve catalog name. Default implementation gets name from CatalogComponent
+     * annotation, if not found, class simpleName is used.
+     */
+    default String getName() {
+        final CatalogComponent annotation = AnnotationUtils.findAnnotation(
+                AopProxyUtils.ultimateTargetClass(this), CatalogComponent.class);
+
+        if (annotation != null) {
+            return annotation.value();
+        } else {
+            return getClass().getSimpleName();
+        }
+    }
+}
--- a/core-api/src/main/java/org/openhubframework/openhub/api/catalog/CatalogComponent.java
+++ b/core-api/src/main/java/org/openhubframework/openhub/api/catalog/CatalogComponent.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.api.catalog;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * Annotation for implementations of {@link Catalog} interface.
+ * It contains catalog name, and does include spring {@link Component} as well.
+ * @author Karel Kovarik
+ * @since 2.0
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Component
+public @interface CatalogComponent {
+
+    /**
+     * The value is catalog name, as it will be available through the Catalog API.
+     *
+     * @return the catalog name.
+     */
+    @AliasFor("name")
+    String value() default "";
+
+    /**
+     * @see #value()
+     */
+    @AliasFor("value")
+    String name() default "";
+}
--- a/core-spi/src/main/java/org/openhubframework/openhub/spi/catalog/CatalogService.java
+++ b/core-spi/src/main/java/org/openhubframework/openhub/spi/catalog/CatalogService.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.spi.catalog;
+
+import java.util.List;
+
+import org.openhubframework.openhub.api.catalog.CatalogEntry;
+
+
+/**
+ * Service to provide OpenHub catalogs.
+ *
+ * @author Karel Kovarik
+ * @since 2.0
+ */
+public interface CatalogService {
+
+    /**
+     * Fetch all entries from catalog identified by name.
+     * @param catalogName the catalog name.
+     * @return list of all entries.
+     */
+    List<CatalogEntry> getEntries(String catalogName);
+}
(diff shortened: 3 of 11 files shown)


core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Contract for OpenHub catalogs. All catalogs that should be available through
+ * catalog API, should be beans implementing this interface.
+ * Note: designed to be used with helper annotation {@link CatalogComponent}.
+ *
+ * @author Karel Kovarik
+ * @see CatalogComponent
+ */
+public interface Catalog {
```
pjuza: Firstly I was little bit confused because I have been using term "codebase", but maybe "catalog" is also good.

core-api/src/main/java/org/openhubframework/openhub/api/catalog/CatalogComponent.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * Annotation to annotate implementation of {@link Catalog} interface.
```
pjuza: "Annotation to annotate" :)

core-spi/src/main/java/org/openhubframework/openhub/spi/catalog/CatalogService.java
```diff
@@ -0,0 +1,21 @@
...
+ * @author Karel Kovarik
+ */
+public interface CatalogService {
+
+    /**
+     * Fetch all entries from catalog identified by name.
+     * @param catalogName the catalog name.
+     * @return list of all entries.
+     */
+    List<CatalogEntry> fetchEntries(String catalogName);
```
pjuza: I don't know if there is some sofisticed logic under "fetchEntries" but we use commonly these conventions:
- find* if can be null return value
- get*, getAll* for getting items

I would use getEntries at this place.

core/src/main/java/org/openhubframework/openhub/core/catalog/CatalogServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+public class CatalogServiceImpl implements CatalogService {
+
+    @Autowired
+    private Optional<List<Catalog>> catalogList;
+
+    @Override
+    public List<CatalogEntry> fetchEntries(final String catalogName) {
+        Assert.hasText(catalogName, "the catalog name must not be empty!");
+
+        return catalogList.flatMap(
```
pjuza: Really nice, I needed to look at to Java documentation to see what it do ... :)

core/src/main/java/org/openhubframework/openhub/core/catalog/SimpleCatalogEntry.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final String code;
+    private final String description;
+
+    public SimpleCatalogEntry(String code, String description) {
+        this.code = code;
+        this.description = description;
+    }
+
+    public String getCode() {
```
pjuza, line 46: Can it be null? If yes then add @Nullable annotation to methods which can return NULL value.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/catalog/rpc/CatalogEntryRpc.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final String code;
+    private final String description;
+
+    public CatalogEntryRpc(String code, String description) {
+        this.code = code;
+        this.description = description;
+    }
+
+    public String getCode() {
```
pjuza, line 45: Again @Nullable add ...

pjuza (approved):
Good job, I like your code.

I don't see any implementation of Catalog interface, any specific catalog implementation.

core-spi/src/main/java/org/openhubframework/openhub/spi/catalog/CatalogService.java
kkovarik: Ok, will rename 👍

core/src/main/java/org/openhubframework/openhub/core/catalog/SimpleCatalogEntry.java
kkovarik: I will make it not null with Asserts on the constructor, it does not make much sense to me to have catalog entry without code.

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
kkovarik: Yeah, it was first name that came up to my mind. Its like everyone uses different terms for this :)

Can rename to codebase, no big deal...

kkovarik:
@pjuza implementations are in messages feature branch:

* first one, for message states: https://github.com/OpenWiseSolutions/openhub-framework/blob/feature/OHFJIRA-57-messages/core/src/main/java/org/openhubframework/openhub/core/catalog/MessageStateCatalog.java
* not sure with the second one though, for source systems: https://github.com/OpenWiseSolutions/openhub-framework/blob/feature/OHFJIRA-57-messages/core/src/main/java/org/openhubframework/openhub/core/catalog/SourceSystemCatalog.java, as there is only interface for source systems, and it must be project specific...

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
pjuza: Catalog is not bad :), it's up to you ...

pjuza:
Thank you for catalog examples, it seems good.

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
hanusto: I add thumb for Catalog

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
```diff
@@ -0,0 +1,36 @@
+package org.openhubframework.openhub.api.catalog;
```
hanusto, line 17: Please add license header - for all classes and "files"

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
hanusto: What do you think about that:

- service interface will be "CatalogService" and annotation will be "Catalog" - from my point of view it will be better than actual naming

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
```diff
@@ -0,0 +1,36 @@
...
+     * Get all entries in catalog.
+     */
+    List<CatalogEntry> getEntries();
+
+    /**
+     * Resolve catalog name. Default implementation gets name from CatalogComponent
+     * annotation, if not found, class simpleName is used.
+     */
+    default String resolveName() {
+        final CatalogComponent annotation = AnnotationUtils.findAnnotation(getClass(), CatalogComponent.class);
```
hanusto: Be carefully if implementation class will by proxied via AOP. Safe variant is:
`AnnotationUtils.findAnnotation(AopProxyUtils.ultimateTargetClass(getClass()), CatalogComponent.class);` but maybe for class resolving it is safe anyway.

core-spi/src/main/java/org/openhubframework/openhub/spi/catalog/CatalogService.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.openhubframework.openhub.api.catalog.CatalogEntry;
+
+
+/**
+ * Service to provide OpenHub catalogs.
+ *
+ * @author Karel Kovarik
+ */
+public interface CatalogService {
```
hanusto: What is difference between `Catalog` and `CatalogService`? Oh, CatalogService is SPI, Catalog is service interface for API.

web-admin/src/main/java/org/openhubframework/openhub/admin/web/catalog/rest/CatalogController.java
```diff
@@ -0,0 +1,45 @@
...
+    private CatalogService catalogService;
+
+    /**
+     * Get full catalog identified by its name.
+     * @param name the name of catalog.
+     * @return list of catalog entries.
+     */
+    @GetMapping(path = "/{name}", produces = {"application/xml", "application/json"})
+    @ResponseStatus(HttpStatus.OK)
+    @ResponseBody
```
hanusto: I think it is not mandatory if RestController annotation is used, right?

web-admin/src/test/java/org/openhubframework/openhub/admin/web/catalog/rest/CatalogControllerTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void getCatalog_noEntriesInCatalog() throws Exception {
+        when(catalogService.getEntries(eq("NOT_EXISTING")))
+                .thenReturn(Collections.emptyList());
+
+        final URIBuilder uriBuilder = createGetUrl(CatalogController.REST_URI + "/NOT_EXISTING");
+
+        // GET /api/catalogs/NOT_EXISTING
```
hanusto: I am thinking about implementation. It is correct that this endpoint returns empty collection for unknown catalog? This catalog does not exist, right? I think that 404 HTTP status code should be returned.

hanusto:
I added some notes. Please @kkovarik check it. And can you please check APIary definition? Sample responses are not correct.

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
kkovarik: Thanks for noticing, added

web-admin/src/test/java/org/openhubframework/openhub/admin/web/catalog/rest/CatalogControllerTest.java
hanusto: @kkovarik please can you explain your solution? This PR should be merged at first before messages.

hanusto (approved):

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
kkovarik: Thanks, sounds reasonable, was not aware of that. Changed to be safe, one never knows when there are layers of proxies :)

web-admin/src/main/java/org/openhubframework/openhub/admin/web/catalog/rest/CatalogController.java
kkovarik: Right, removed.

web-admin/src/test/java/org/openhubframework/openhub/admin/web/catalog/rest/CatalogControllerTest.java
kkovarik: Fixed directly in controller, unit test updated. Now it should be compliand with the apiary. 404 for not existing catalog.

core-api/src/main/java/org/openhubframework/openhub/api/catalog/Catalog.java
kkovarik: Annotation could be Catalog, agreed, however CatalogService is already the SPI interface, which will be used whenever you want some catalog value, co I tried to make it intuitive - "CatalogService".

So 3 names are necessary:
* annotation - currently CatalogComponent (motivation for name is it does contain spring Component within, so it is clear one does not need to add it) -> Catalog?
* interface for single catalog - currently Catalog -> CatalogService?
* service in SPI to provide access to all catalogs - CatalogService -> ?

PR #64, opened 2017-07-17
Fixed problem with starting OpenHub as standalone application

hanusto:
Explicitly mark environment (application) as web-based and avoid to exception: `@ConditionalOnWebApplication` (required) not a web application.

Issue: NOJIRA


hanusto:
It is possible to test this in Tomcat deployment? @kkovarik or @pjuza ? Thx.

kkovarik:
@hanusto great, tested on local tomcat, seems to be working ok, spring context loads fine.

hanusto:
One more known issue for standalone - non visible WSDL services. It is necessary to be fixed.

pjuza:
I have also tested on local tomcat, it works fine.
Web services are working but OHF started without admin console module.

kkovarik (approved):

PR #67, opened 2017-08-10
OHFJIRA-66 springboot jackson cors and pagination

hanusto:
- Removed specific Jackson configuration, pageable custom definition.
- Improved schema script for postgreSQL migration
- Change DB on Heroku - H2 -> PostgreSQL.

Issue: [OHFJIRA-66](https://openhubframework.atlassian.net/projects/OHFJIRA/issues/OHFJIRA-66?)


kkovarik (approved):

core-api/pom.xml
```diff
@@ -100,6 +100,11 @@
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-spring-security</artifactId>
         </dependency>
+        <!-- we use Jackson annotations in API to define global rules of serialization/deserialization -->
```
hanusto: @pjuza @kkovarik Actually we have bug how to serialize object represented by interface, for example ExternalSystemExtEnum. This solution is perfect, because it is in one place, but I have problem to deploy application (with this dependency in CORE-API) to heroku. I dont know why, but question is how to fix this bug. For example we can update getter methods in MessageBaseRpc to String return value. Have you any idea?

PR #68, opened 2017-08-15
Feature/messages improvements

kkovarik:
#### CHANGES
* added customData field to the detail of message (added to apiary as well)
* fix of possible NPE for not mandatory enums in rpc
* synced states for RESTART & CANCEL of messages

### DIFF
--- a/web-admin/src/main/java/org/openhubframework/openhub/admin/web/message/rpc/MessageBaseRpc.java
+++ b/web-admin/src/main/java/org/openhubframework/openhub/admin/web/message/rpc/MessageBaseRpc.java
@@ -29,6 +29,8 @@ import org.openhubframework.openhub.api.entity.ServiceExtEnum;
 import org.openhubframework.openhub.api.exception.ErrorExtEnum;
 import org.springframework.core.convert.converter.Converter;
 
+import javax.annotation.Nullable;
+
 
 /**
  * Base RPC for message entities.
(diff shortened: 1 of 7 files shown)


web-admin/src/main/java/org/openhubframework/openhub/admin/web/message/rpc/MessageBaseRpc.java
```diff
@@ -89,7 +89,7 @@ public void setState(MsgStateEnum state) {
     }
 
     public String getErrorCode() {
-        return errorCode.getErrorCode();
+        return errorCode != null ? errorCode.getErrorCode() : null;
```
hanusto: Yes, I prepared bugfix branche https://github.com/OpenWiseSolutions/openhub-framework/compare/feature/bugfix with the same bugfix.

hanusto (approved):

web-admin/src/main/java/org/openhubframework/openhub/admin/web/message/rpc/MessageBaseRpc.java
kkovarik: Ok :) Was not able to end-to-end test it without fixing it...

hanusto: Sure.

PR #69, opened 2017-08-29
Config params : improved data types formating

kkovarik:
### ISSUE
* db params are stored in db as Strings, on RPC the same, therefore when jackson marshalled them they were always with quotes - "true", "42"

### CHANGES
* in DbConfigurationParamRpc, currentValue & defaultValue of param is now java.lang.Object. When DB entity is converted, there is extra manual serialization & deserialization from db entity String.
* DataTypeEnum was extended for each type it now has defined two Functions. one for serialization (Object -> String) other one inverse (String -> Object).
* for example, integer parameter: in DB it is stored still as string, when DB entity is converted to DbConfigurationParamRpc, function Integer.parse() defined in DataTypeEnum is applied

### DIFF
--- a/core-api/src/main/java/org/openhubframework/openhub/api/configuration/DataTypeEnum.java
+++ b/core-api/src/main/java/org/openhubframework/openhub/api/configuration/DataTypeEnum.java
@@ -21,6 +21,9 @@ import java.io.Serializable;
 import java.time.LocalDate;
 import java.util.regex.Pattern;
 
+import org.openhubframework.openhub.common.converter.DataTypeConverter;
+import org.openhubframework.openhub.common.converter.TypeConverterEnum;
+
 
 /**
  * Enumeration of possible data types of configuration parameters.
@@ -33,42 +36,45 @@ public enum DataTypeEnum {
     /**
      * Text, string (default).
      */
-    STRING(String.class),
+    STRING(String.class, TypeConverterEnum.STRING),
 
     /**
      * Integer, whole number.
      */
-    INT(Integer.class),
+    INT(Integer.class, TypeConverterEnum.INT),
 
     /**
      * Floating number.
      */
-    FLOAT(Float.class),
+    FLOAT(Float.class, TypeConverterEnum.FLOAT),
 
     /**
      * Date (without time).
      */
-    DATE(LocalDate.class),
+    DATE(LocalDate.class, TypeConverterEnum.DATE),
 
     /**
      * Boolean.
      */
-    BOOLEAN(Boolean.class),
+    BOOLEAN(Boolean.class, TypeConverterEnum.BOOLEAN),
 
     /**
      * File.
      */
-    FILE(File.class),
+    FILE(File.class, TypeConverterEnum.FILE),
 
     /**
      * Regular pattern.
      */
-    PATTERN(Pattern.class);
+    PATTERN(Pattern.class, TypeConverterEnum.PATTERN);
 
     private Class<?> typeClass;
 
-    DataTypeEnum(Class<?> typeClass) {
+    private DataTypeConverter converter;
+
+    DataTypeEnum(Class<?> typeClass, DataTypeConverter converter) {
         this.typeClass = typeClass;
+        this.converter = converter;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


core-api/src/main/java/org/openhubframework/openhub/api/configuration/DataTypeEnum.java
```diff
@@ -33,42 +35,48 @@
     /**
      * Text, string (default).
      */
-    STRING(String.class),
+    STRING(String.class, stringDeserialization(), stringSerialization()),
```
pjuza: Nice solution via JDK8 functions but please two things for considaration:
- there is org.openhubframework.openhub.common.converter.TypeConverterEnum - use this class or/and replace implementation by your implementation? Your implementation is identical to TypeConverterEnum besides from FILE and PATTERN
- why don't use converters in Spring (org.springframework.core.convert.ConversionService), why implement something by ourselves?

pjuza (changes requested):
I don't know exactly how to solve it but we have duplicity at this moment.

Maybe question is if this conversion should be in core module (now DataTypeEnum) and not in web module because it's specific for REST services ...

kkovarik:
Update: now conversion is done directly on db entity - getCurrentValueRaw & getDefaultValueRaw methods added...

core-api/src/main/java/org/openhubframework/openhub/api/configuration/DataTypeEnum.java
kkovarik: Well, tried solution with ConversionService, however does not seem nice solution without bigger refactoring, as conversion is done on db entity (or rpc). So I would have to pass the conversion service bean somehow. Via getters or setters or other methods.

So now I went for the TypeConverterEnum as simple solution, please check what it looks like now.

core-api/src/main/java/org/openhubframework/openhub/api/configuration/DbConfigurationParam.java
```diff
@@ -183,6 +184,72 @@ public String getValue() {
     }
 
     /**
+     * Gets current value converted to the object of appropriate type (integer, boolean etc).
+     *
+     * @return the current raw value.
+     */
+    public Object getCurrentValueRaw() {
```
pjuza: I don't know if "raw" value is the best name. What about getCurrentValueAsObject or change original method getCurrentValue->getCurrentValueAsString and use getCurrentValue for this.

web-admin/src/test/java/org/openhubframework/openhub/admin/web/configuration/rest/DbConfigurationControllerTest.java
```diff
@@ -102,8 +102,8 @@ public void testUpdate() throws Exception {
                 .add("code", param.getCode())
                 .add("categoryCode", param.getCategoryCode())
                 .add("dataType", param.getDataType().name())
-                .add("currentValue", "222")
-                .add("defaultValue", "10")
+                .add("currentValue", 222)
+                .add("defaultValue", 10)
```
pjuza, line 156: Unit test for boolean would be great.

pjuza (approved):
Nice solution ...

core-api/src/main/java/org/openhubframework/openhub/api/configuration/DbConfigurationParam.java
```diff
@@ -183,6 +184,72 @@ public String getValue() {
...
+        }
+    }
+
+    /**
+     * Sets default value from object of appropriate type directly, it will be converted to string
+     * to be stored in repository.
+     *
+     * @param defaultValue the raw object of default value.
+     */
+    public void setDefaultValueRaw(Object defaultValue) {
```
hanusto: Change the name of setter as well.

web-admin/src/test/java/org/openhubframework/openhub/admin/web/configuration/rest/DbConfigurationControllerTest.java
hanusto: I agree

hanusto (changes requested):
Please rename methods and add test to verify correct behaviour for boolean based types.

web-admin/src/test/java/org/openhubframework/openhub/admin/web/configuration/rest/DbConfigurationControllerTest.java
kkovarik: Added unit test for get & update of boolean attributes, good idea 👍

kkovarik:
Renamed getters & setters, thanks for feedback!

hanusto (approved):
Thanks for PR.

PR #70, opened 2017-09-08
Feature/integration test config

kkovarik:
Changed configuration to enable integration tests, see: https://github.com/OpenWiseSolutions/openhub-ri/pull/1

### CHANGES
* MvcPageableConfiguration removed from spring.factories. Now, when openhub-web module is referenced, this file is picked up and therefore beans from packages like org.openhubframework.openhub.admin are needed or spring context will not start. Switched to "normal" Configuration class.
* ~~added dependency to tomcat to test module, for integration tests to have embedded servlet container ready.~~
* ~~added TestWebConfig class with prepared autoscan for openhub-web beans~~

### DIFF
--- a/web/src/main/java/org/openhubframework/openhub/web/config/MvcPageableConfiguration.java
+++ b/web/src/main/java/org/openhubframework/openhub/web/config/MvcPageableConfiguration.java
@@ -19,13 +19,13 @@ package org.openhubframework.openhub.web.config;
 import javax.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
 import org.springframework.data.web.SortHandlerMethodArgumentResolver;
 
 import org.openhubframework.openhub.admin.web.common.rpc.paging.OffsetBasedPagingProperties;
-import org.openhubframework.openhub.common.AutoConfiguration;
-
 
 /**
  * Spring configuration to pageable support.
@@ -34,7 +34,8 @@ import org.openhubframework.openhub.common.AutoConfiguration;
  * @since 2.0
  * @see org.springframework.context.support.PropertySourcesPlaceholderConfigurer
  */
-@AutoConfiguration
+@ConditionalOnWebApplication
+@Configuration
 public class MvcPageableConfiguration {
 
     @Autowired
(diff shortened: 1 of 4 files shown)


test/src/main/java/org/openhubframework/openhub/test/TestWebConfig.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Configuration for tests that require entire web context.
+ * Does configure component scan to include packages needed for web context initialization.
+ *
+ * @author Karel Kovarik
+ * @since 2.0
+ */
+@ComponentScan(basePackages = {
+        "org.openhubframework.openhub.web",
```
pjuza: Previously was org.openhubframework.openhub.web.common, isn't it misspelling?

test/pom.xml
```diff
@@ -75,6 +75,11 @@
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-commons</artifactId>
         </dependency>
+        <!-- added tomcat to enable web tests, exclude when needed -->
+        <dependency>
```
pjuza: Is it necessary? I wanted to keep test module as lighweight as possible, but if has to be ...

pjuza (dismissed):

test/pom.xml
kkovarik: No, it is only necessary for integration tests, will remove...

test/src/main/java/org/openhubframework/openhub/test/TestWebConfig.java
kkovarik: No. There is more settings that I need for integration tests - like security, that is configured in another subpackage under org.openhubframework.openhub.web.config...so I added the parent...

hanusto: Question is whether we should support web-based integration tests via `openhub-test`. Last result is for now NO. So it is correct to define `TestWebConfig` in openhub-test? Maybe compromise is moved this class to RI project where standard project specific classes will be created via archetype.

test/pom.xml
hanusto: Remove

test/src/main/java/org/openhubframework/openhub/test/TestWebConfig.java
hanusto: BTW: test support dependency for internal use, core of framework, is something other than support for testing project implementation. In tests to verify project implementation user probably wants to use dependency to core framework functions, but it is cyclic dependency, where test artifact will have dependency to core (to use classes) and core has dependency to test for testing support.

hanusto:
It is necessary to resolve conflicts.

kkovarik:
@hanusto @pjuza Rebased on top of current develop.

Changes from previous iteration:
- removed dependency to tomcat & TestWebConfig class as it will be solved in openhub-ri

hanusto (approved):

PR #71, opened 2017-09-08
Fixed update payload for config params

bionaut:
@kkovarik found minor payload bug
**before** > INTs and FLOATs were send as strings to BE
**now** > INT as INT, FLOAT as FLOAT


kkovarik (approved):
Working fine, thanks!

hanusto (approved):

PR #73, opened 2017-09-11
OHFJIRA-22: project implementation support

hanusto:
Added support to simply create project implementation based upon OpenHub. For that new Reference implementation of OpenHub created - https://github.com/OpenWiseSolutions/openhub-ri.

Changes:
- Added support for external configuration files, for example to override openhub core property values.
- Bugfix: Project implementation based upon OpenHub - repair error in saved node_id in Message entity into DB
- Refactored web module. Now it is jar which can be referenced. To build OpenHub standalone application, new module war is created.

Issue: [OHFJIRA-22](https://openhubframework.atlassian.net/browse/OHFJIRA-22)


kkovarik (approved):
Seems fine to me.

PR #74, opened 2017-09-11
Preparation for candidate release

hanusto:
Actual version of OpenHub is not backward compatible with previous version, we have to increase MAJOR version (http://semver.org/) before releasing. This PR has to be last PR before releasing candidate.

Issue: NOJIRA


pjuza (approved):
Will be SNAPSHOT version for RC? Or is possible to have "rc" directly in the version name? 
We can communication that snapshot version is release candidate version but it is not standard.

kkovarik (approved):

PR #77, opened 2018-02-28
NOJIRA: RequestSending & ResponseReceive EventNotifiers config

kkovarik:
### ISSUE
* when **ohf.requestSaving.enable** is set to true (and request saving enabled), it does fail on loading the field **ConfigurationItem&lt;Pattern&gt; endpointFilterPattern** as it cannot convert the String from DB to Pattern.
* solved with setting the property as String and compile Pattern afterward "manually".
 _Tried the possibility of adding String -> Pattern converter to Spring ConversionService, however it is not as simple as is seems. I just wanted to register spring converter, to be able to make this conversion. So I said, ok, I will enrich the spring ConversionService bean to support it. Pitfall: Properties from the database are loaded into spring properties. And propertyResolver does use its own ConversionService, not the bean in context.
see: https://github.com/spring-projects/spring-framework/blob/4.3.x/spring-core/src/main/java/org/springframework/core/env/AbstractPropertyResolver.java_

### FOR DISCUSSION
* **do we want the pattern to be always loaded "fresh" in runtime**? even if it means that every time event is intercepted, it will load the property from DB and compile regex.Pattern from it? I was kind of afraid of performance, so I prepared the variant where only once (on startup) the pattern is loaded from DB & compiled.
* Note: Tried to realize this the "standard" way - implementing InitiazingBean or use PostConstruct annotation, however it did not worked. Method was always invoked sooner, than any of DB properties were loaded. Therefore I had to hook this "pattern initialization" to spring event - ApplicationReady, ContextRefreshed worked as well.

### DIFF
--- a/core/src/main/java/org/openhubframework/openhub/core/reqres/RequestSendingEventNotifier.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/reqres/RequestSendingEventNotifier.java
@@ -28,6 +28,8 @@ import org.apache.camel.management.event.ExchangeSendingEvent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
+import org.springframework.context.event.EventListener;
 import org.springframework.util.Assert;
 
 import org.openhubframework.openhub.api.asynch.AsynchConstants;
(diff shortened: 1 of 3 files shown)


hanusto:
Thank you Karel. Changes are OK. I created [OHFJIRA-82](https://openhubframework.atlassian.net/projects/OHFJIRA/issues/OHFJIRA-82) issue into backlog. We can reopen this question in next development life cycle.

hanusto (approved):

pjuza:
A added OHFJIRA-83 - notify admin user if change of some property needs to restart the server.

PR #79, opened 2018-03-04
Merge bugfixes from R2.0

hanusto:
## Changes:

- relative web context path fixed
- `WsdlController` respect contextPath on which the openhub is deployed
- javaMelody fixed configuration, when it is already initialized by servlet container
- admin console external link for javamelody changed to support relative context
- `RequestSending` & `ResponseReceive` EventNotifiers - endpoint pattern switch to String
- For community reason we change slogan for better understanding what OpenHub is and does


kkovarik (commented):
Seems fine 👍

kkovarik (approved):

PR #89, opened 2018-11-05
[OHFJIRA-84] : java 11 runtime compatibility

kkovarik:
## Support for JDK 11 runtime

With these added libraries, openhub is able to run with java 11 runtime.

### Details
Main issue with backward compatibility is removal of some Java EE APIs:
https://cr.openjdk.java.net/~iris/se/11/spec/fr/java-se-11-fr-spec/#APIs-removed

these apis were marked as deprecated in java 9, removed entirely in java 11.
https://stackoverflow.com/questions/43574426/how-to-resolve-java-lang-noclassdeffounderror-javax-xml-bind-jaxbexception-in-j

Following dependencies were added, as they are required for openhub library stack:
```
javax.xml.bind:jaxb-api
com.sun.xml.bind:jaxb-core
com.sun.xml.bind:jaxb-impl
javax.xml.ws:jaxws-api
com.sun.xml.messaging.saaj:saaj-impl
com.sun.activation:javax.activation
```
Versions used are the ones used in java 8 runtime (when possible).

#### Warning during spring context startup:
```
matching: class path resource [sp_h2_server.xml]
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.springframework.cglib.core.ReflectUtils$1 (jar:file:/home/app.war!/WEB-INF/lib/spring-core-4.3.20.RELEASE.jar!/) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of org.springframework.cglib.core.ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
Caused by spring framework, does seem to be warning only, **will be resolved in Spring 5.1** (https://jira.spring.io/browse/SPR-15859). Warning should be resolved with migration to Spring Boot 2.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,6 +206,11 @@
         <javamelody-version>1.65.0</javamelody-version>
         <metrics-version>3.2.0</metrics-version>
         <jolokia-version>1.3.5</jolokia-version>
+        <!-- java ee -->
+        <jaxb.version>2.2.11</jaxb.version>
+        <jaxws.version>2.2.12</jaxws.version>
+        <javax-activation.version>1.2.0</javax-activation.version>
+        <saaj-impl.version>1.4.0</saaj-impl.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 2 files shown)


pjuza:
I have Java 8 and Tomcat 8.0.20 and I'm not able to start application server - the following error occurs:

> 06-Nov-2018 22:31:26.961 SEVERE [RMI TCP Connection(2)-127.0.0.1] org.apache.catalina.startup.ContextConfig.processAnnotationsJar Unable to process Jar entry [module-info.class] from Jar [jar:file:/Volumes/Obelix/projects/openhub-framework/war/target/openhub/WEB-INF/lib/jaxws-api-2.3.0.jar!/] for annotations
 org.apache.tomcat.util.bcel.classfile.ClassFormatException: Invalid byte tag in constant pool: 19

It's caused by updating the JAX-WS libraries of my current project to version 2.3.0, see https://blog.doubleslash.de/unable-to-process-jar-entry-module-info-class-from-jar/

Maybe it's time to leave Java 8 (also Java 9 has reached end of support) and move forward to Java 10 and higher but we have to change documentation at least (https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33608/Getting+started)

hanusto:
But it is incompatible change. We should ship that in next major version. Why we increase version of Jax-libs? It is not possible to use the same included in Java 8?

kkovarik:
It is not issue of java 8 runtime itself, it is caused by the tomcat 8.x. I have changed versions of libraries, should improve compatibility with older versions.

pjuza:
I have tested with Tomcat 8.0 and 8.5 and it works fine. Great 👍

pjuza (approved):

PR #90, opened 2018-11-05
[OHFJIRA-98] : registration of TestWsUriBuilder bind to property

kkovarik:
### Issue
Now unable to create integration test, with "live" webservice call.

### Changes
* changed registration of TestWsUriBuilder bean to the spring context, now it can be controlled via property 
* by default it is enabled, even if property is missing, to be backward-compatible as much as possible

### Documentation
* is there already any info on confluence about TestWsUriBuilder? if not and PR is approved, I will add some mention and that it can be disabled/enabled here: https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/8814613/How+to+write+unit+test

### DIFF
--- a/examples/src/test/java/org/openhubframework/openhub/modules/AbstractExampleModuleTest.java
+++ b/examples/src/test/java/org/openhubframework/openhub/modules/AbstractExampleModuleTest.java
@@ -16,6 +16,7 @@
 
 package org.openhubframework.openhub.modules;
 
+import org.openhubframework.openhub.test.route.EnableTestWsUriBuilder;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.ContextConfiguration;
 
@@ -30,6 +31,7 @@ import org.openhubframework.openhub.test.route.ActiveRoutes;
  *
  * @author Petr Juza
  */
+@EnableTestWsUriBuilder
 @ActiveRoutes(classes = ExceptionTranslationRoute.class)
 @ContextConfiguration(classes = ExampleTestConfig.class)
 @ActiveProfiles(profiles = ExampleProperties.EXAMPLE_PROFILE)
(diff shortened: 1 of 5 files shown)


pjuza:
Comment to wiki - mentioned wiki page is correct but it's not only about adding new info but also correction of current examples, if necessary. I havn't found any other place so far ...

hanusto:
How programmer can solve a situation where he want to write both tests - integration with real time execution and as well as mocking?

hanusto (approved):

pjuza:
I have similar doubt as Tomas - wouldn't be better to solve it in source code only, without using configuration parameter? What about use annotation in combination with @Conditional (e.g. https://blog.andresteingress.com/2017/09/21/spring-boot-conditional-annotations.html)?

    ActiveRoutes(classes = SyncHelloRoute.class)
    @TestWsBuilder
    public class SyncHelloRouteTest extends AbstractExampleModuleTest {

    @Primary
    @Bean
    @Conditional(...)
    public WebServiceUriBuilder testWebServiceUriBuilder() {
        return new TestWsUriBuilder();
    }

kkovarik:
To answer Tomas question: Well, I expect he would write them to separate test classes. One extending something like "AbstractIntegrationTest" the other one like "AbstractMockedTest". The parents would setup configuration, either via `TestPropertySource` directly, or with application properties with its own profile.

The most important is, that you have the ability to replace the TestWsUriBuilder with your own bean now. With Conditional on property it seemed like the simplest solution. What condition do you have in mind, @pjuza ?

pjuza:
We have decided to use annotation @EnableTestWsUriBuilder instead of property configuration. By default new bean TestWsUriBuilder won't be created unless annotation is used on class level.

The change is not backward-compatible => we will decribe it into release notes.

kkovarik:
PR updated with new annotation-controlled way.

pjuza (approved):

PR #92, opened 2018-11-05
[OHFJIRA-97] : hazelcast discovery of other nodes disabled by default

kkovarik:
Configuration of hazelcast, to disable multicast by default.

### CHANGES
Disabled all discoveries, hazelcast is in standalone mode. Cluster should be configured explicitely, with proper discovery, which may, or may not be multicast.

### DOCUMENTATION
Hazelcast page on confluence updated:
https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/11111718/Hazelcast

### DIFF
--- a/core/src/main/resources/config/ohf_hazelcast.xml
+++ b/core/src/main/resources/config/ohf_hazelcast.xml
@@ -21,7 +21,7 @@
     See the following link for inspiration/documentation:
     https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/resources/hazelcast-full-example.xml
 -->
-<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.5.xsd"
+<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.7.xsd"
            xmlns="http://www.hazelcast.com/schema/config"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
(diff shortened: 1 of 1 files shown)


core/src/main/resources/config/ohf_hazelcast.xml
```diff
@@ -31,19 +31,12 @@
     </group>
     <management-center enabled="false" update-interval="10">http://localhost:8080/mancenter</management-center>
```
hanusto, line 32: Maybe we should document it in wiki (http://localhost:8080/mancenter).

hanusto (approved):

core/src/main/resources/config/ohf_hazelcast.xml
```diff
@@ -31,19 +31,12 @@
...
         <join>
-            <multicast enabled="true">
-                <multicast-group>224.2.2.3</multicast-group>
-                <multicast-port>54327</multicast-port>
-            </multicast>
+            <!-- by default cluster discovery is not configured -->
+            <multicast enabled="false" />
+            <tcp-ip enabled="false" />
+            <aws enabled="false" />
+            <discovery-strategies />
```
pjuza: Is any reason why you add <discovery-strategies />? It serves for defining custom discovery strategy (https://docs.hazelcast.org/docs/latest-development/manual/html/Extending_Hazelcast/Discovery_SPI/Discovery_Strategy.html) and I don't think it will be something common ...

pjuza (approved):

core/src/main/resources/config/ohf_hazelcast.xml
kkovarik: Well, idea was to list config params, disabled. In this case empty ... just to enumerate all of config parameters. Current version of Hazelcast for example, has more join methods available - like k8s, eureka...

However I´m no expert in Hazelcast configuration, so if it is not common, I will remove it...:)

core/src/main/resources/config/ohf_hazelcast.xml
kkovarik: Added short mention about it to the wiki

core/src/main/resources/config/ohf_hazelcast.xml
kkovarik: Removed discovery-strategies

PR #93, opened 2018-11-07
[OHFJIRA-95] Custom datasource support

kkovarik:
## Custom datasource support

This PR should server as discussion base for possibilities on how to add custom persistence to OpenHub-based application. I shall then add these possibilities to the confluence, when it is approved etc.

How can you save something, preferrably by using JPA?

### 1. Add some table directly to the openhub database.
- imho it can make sense in some straight-forward scenarios, when I know I will only need one more table
- see first commit in this PR, it adds simple customization to the EntityManagerFactoryBuilder - specification of so called **"additional packages"**. These additional packages are then registered to the entityManager. Then custom dao can be built or spring-data-jpa repositories can be configured for these additional entities as well.
Configuration: 
`ohf.jpa.additional-packages=org.openhubframework.openhub.ri.persistence`

### 2. Add second datasource, without JPA
- no issue in this setup, except that OpenHubDatasource is Primary bean, as long as project is working with Qualified dataSource, its ok. JdbcTemplate can be used, or entirelly wrapped service (not exposing beans to spring context).

### 3. Add second datasource, with global transaction (one transactionManager)
- SpringBoot autoconfiguration can be levered, either setup with JTA (Atomikos, Bitronix) or provided by application server
- tested on OpenHub-ri with JTA Bitronix, one dataSource for PostgreSQL, second for MySQL, JpaRepositories on custom tables
...
### 4. Add second datasource, each having its own transactionManager
* now it could not be achieved, as openHub transactionManager is configured entirelly by Spring Boot, which does not auto configure transactionManager if another is already present
* see second commit in this PR - added therefore possibility to create transactionManager explicitely for the OpenHub datasource. It will be instance of https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/orm/jpa/JpaTransactionManager.html
* configuration: `ohf.jpa.transaction-manager.enabled` , by default it is **disabled**
* setup can look like this:
```
@EnableJpaRepositories(
        entityManagerFactoryRef = "customEntityManagerFactory",
        transactionManagerRef = "customTransactionManager",
        basePackages = "org.openhubframework.openhub.ri.persistence.repository")
@Configuration
public class SecondDbConfiguration {

    @Bean
    @Qualifier("customTransactionManager")
    PlatformTransactionManager customTransactionManager() {
        return new JpaTransactionManager(customEntityManagerFactory().getObject());
    }

    @Bean
    @Qualifier("customEntityManagerFactory")
    LocalContainerEntityManagerFactoryBean customEntityManagerFactory() {
        HibernateJpaVendorAdapter jpaVendorAdapter = new HibernateJpaVendorAdapter();
        jpaVendorAdapter.setGenerateDdl(false);

        LocalContainerEntityManagerFactoryBean factoryBean = new LocalContainerEntityManagerFactoryBean();

        factoryBean.setJpaVendorAdapter(jpaVendorAdapter);
        factoryBean.setPackagesToScan(User.class.getPackage().getName());
        factoryBean.setDataSource(customDataSource());
        return factoryBean;
    }

    @Bean
    @Qualifier("customDataSource")
    DataSource customDataSource() {
        return DataSourceBuilder.create()
                                .url("jdbc:mysql://localhost:3306/testdb?useSSL=false&nullNamePatternMatchesAll=true")
                                .driverClassName("com.mysql.jdbc.Driver")
                                .username("root")
                                .password("*****")
                                .build();
    }
```
Please note, that dataSource & entityManagerFactory & transactionManager in OpenHub are Primary beans, therefore in project, Qualifiers must be used.

### Final note
* approaches 3 vs 4 are somewhat in opposite. First idea was to use Qualified beans in OpenHub, not to interfere with project implementation. However it would lead not to be able to leverage Spring autoconfiguration. Imho it´s something we do not want, as it is quite complex. That´s the reason for this approach.


core/src/main/java/org/openhubframework/openhub/core/config/JpaConfigurationProperties.java
```diff
@@ -0,0 +1,48 @@
+package org.openhubframework.openhub.core.config;
```
hanusto, line 1: License header.

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
```diff
@@ -37,16 +47,47 @@
...
+     * Configuration properties instance.
+     */
+    @Autowired
+    private JpaConfigurationProperties jpaConfigurationProperties;
+
     /**
      * Configures JPA entity manager.
      */
     @Bean
+    @Primary
```
hanusto, line 62: Is there possibility that someone already created own LocalContainerEntityManagerFactoryBean? There is no another option how solve that?

hanusto: Maybe somebody can exclude JpaConfig from autoconfiguration via spring.factories to full control.

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
```diff
@@ -37,16 +47,47 @@
...
     public LocalContainerEntityManagerFactoryBean entityManagerFactory(EntityManagerFactoryBuilder builder,
-            DataSource dataSource) {
+               @OpenHubDataSource DataSource dataSource) {
```
hanusto: Why not only @OpenHub ?

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
```diff
@@ -37,16 +49,69 @@
...
+     * @param builder the EntityManagerFactoryBuilder.
+     * @param dataSource the OpenHub dataSource.
+     * @return instance of transactionManager.
+     * @since 2.1
+     * @see JpaTransactionManager
+     */
+    @ConditionalOnProperty(name = JpaConfigurationProperties.TRANSACTION_MANAGER_ENABLED)
+    @Primary
+    @Bean
+    public PlatformTransactionManager transactionManager(EntityManagerFactoryBuilder builder,
```
hanusto, line 94: But how user can define two transaction managers? Because this one has no specific qualifier? There is no error "duplicity bean"?

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfigurationProperties.java
kkovarik: Thanks, will add

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
kkovarik: As OpenHubDataSource is already present. However can refactor...

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
kkovarik: There is no error "duplicity bean" if one of them is Primary. That is how the OpenHubDataSource is configured now.

User will just define another one, with Qualifier, and will use it in his own code. This primary will be used otherwise (in OpenHub core)

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
kkovarik: Basic concept is inherited from OpenhubDatasource, which is Primary bean already. So imho EntityManagerFactoryBean should be as well, to keep things consistent.

I guess there is none win-win scenario. 
First option (as is state):
when you want to create your own entityManagerFactoryBean, you have to make it Primary

Second option:
core openHub is already primary, so you have to make yours Qualified and use it strictly with Qualifier.

core/src/main/java/org/openhubframework/openhub/core/config/JpaConfig.java
kkovarik: Switched to new OpenHub qualifier annotation

kkovarik:
Documentation on wiki updated : https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/782958609/How+to+add+additional+datasource

pjuza (approved):

PR #94, opened 2018-11-09
[OHFJIRA-6] : messages in final state

kkovarik:
## Final messages handling
**Motivation**: to have mechanism how to handle old messages in final state, that would otherwise be in datastore forever.

**Implemented solution**: job is deleting messages after some preconfigured time has elapsed from their last update time.

### Implementation
tried to keep it reasonably simple, yet customizable. There is job, service and collection of FinalMessageProcessors:

#### FinalMessagesProcessingJob
* job, configured to be executed only on one node of cluster
* does call FinalMessagesProcessingService

#### FinalMessagesProcessingService
* finds messages in db that are eligible for "processing" (aka delete)
* provided implementation does use configuration properties for each state
* states are hardcoded in current implementation, there is check that all parameters are configured
* does invoke all FinalMessageProcessors on each message, catches exceptions so that failure to delete one message should not affect the others
* all the processors are invoked in single transaction

#### interface FinalMessageProcessor
* all beans of this type will be applied to message
* order is defined expicitely & respected

#### DeleteFinalMessageProcessor implements FinalMessageProcessor
* processor, enabled/disabled by property
* does delete provided message with all the related entities - extcalls & requests & responses.
 
#### AbstractFinalMessageProcessor
* abstract class to ease FinalMessageProcessor implementation

#### Notes
* had to add some delete methods to Dao´s, did not added them to Service layer, as they are only called from DeleteMessageProcessor
* job and all the beans are conditionaly registered by using property `ohf.asynch.finalMessages.processingEnabled`, by default disabled
* minor backward compatible change in MessageFilter, to enable search for messages by state & lastUpdate only, not to have to add receivedFrom

### Configuration
```
# final messages processing enabled or disabled. Note: cannot be defined in database, 
# needs to be in properties.
ohf.asynch.finalMessages.processingEnabled = true/false

# final messages processing job interval, in seconds.
ohf.asynch.finalMessages.processingIntervalSec=

# maximum number of messages processed per job.
ohf.asynch.finalMessages.iterationMessageLimit=10000

# duration in seconds to keep messages in OK state in the datastore. After this period expires and
# there is no other action with the message, it will be processed as final (deleted probably)
ohf.asynch.finalMessages.ok.saveTimeInSec=0

# duration in seconds to keep messages in CANCEL state in the datastore.
ohf.asynch.finalMessages.cancel.saveTimeInSec=28800

# duration in seconds to keep messages in FAILED state in the datastore.
ohf.asynch.finalMessages.failed.saveTimeInSec=14400

# enable/disable processor that does delete messages
ofh.asynch.finalMessages.deleteProcessor.enabled=true
```
### Possible pittfals 
* performance - how many messages should be processed in single iteration? when enabled on already full database? now there is configurable limit. however it is fetched by single query, could make sense to make multiple queries if this parameter should be too big...

### DIFF
--- a/core-api/src/main/java/org/openhubframework/openhub/api/asynch/finalmessage/FinalMessageProcessor.java
+++ b/core-api/src/main/java/org/openhubframework/openhub/api/asynch/finalmessage/FinalMessageProcessor.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.api.asynch.finalmessage;
+
+import org.openhubframework.openhub.api.entity.Message;
+import org.springframework.core.Ordered;
+
+/**
+ * Contract for final messages processor.
+ * Messages that are no-longer considered relevant for openhub async workflow are expected
+ * to be processed by this processor.
+ *
+ * There can be multiple processors, that are invoked in order defined in getOrder method.
+ *
+ * All processor are expected to be executed in transaction, if any should throw
+ * an exception, transaction will be rolled back.
+ *
+ * @author Karel Kovarik
+ * @since 2.1
+ */
+@FunctionalInterface
+public interface FinalMessageProcessor extends Ordered {
+
+    /**
+     * Default order, if {@link FinalMessageProcessor#getOrder()} method is not implemented.
+     */
+    int DEFAULT_ORDER = 0;
+
+    /**
+     * Process message in OpenHub datastore.
+     *
+     * @param message the message to be processed.
+     */
+    void processMessage(Message message);
+
+    /**
+     * Get the order for processor.
+     *
+     * @return the order of processor, lower value is expected to be invoked first.
+     */
+    default int getOrder() {
+        return DEFAULT_ORDER;
+    }
+}
--- a/core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
+++ b/core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
@@ -83,6 +83,44 @@ public class CoreProps {
     public static final String ASYNCH_POSTPONED_INTERVAL_WHEN_FAILED_SEC = PREFIX + "asynch.postponedIntervalWhenFailedSec";
 
     /**
+     * Final messages processing enabled or disabled. Note: cannot be defined in database, needs to be in properties.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_PROCESSING_ENABLED = PREFIX + "asynch.finalMessages.processingEnabled";
+
+    /**
+     * Final messages processing job interval, in seconds.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_PROCESSING_INTERVAL_SEC = PREFIX + "asynch.finalMessages.processingIntervalSec";
+
+    /**
+     * Maximum number of messages processed per job.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_ITERATION_MESSAGE_LIMIT = PREFIX + "asynch.finalMessages.iterationMessageLimit";
+
+    /**
+     * Configuration of final message processor that does delete messages from the datastore.
+     * Note: final message processing needs to be enabled in order to do that.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_DELETE_PROCESSOR_ENABLED = PREFIX + "asynch.finalMessages.deleteProcessor.enabled";
+
+    /**
+     * Prefix for setting duration to keep messages in final states in the datastore.
+     * After this period expires and there is no other action with the message, it will be processed as final (deleted probably).
+     * Setting to '0' means, that messages will be processed as soon as possible (next scheduled job).
+     * Setting to '-1' means, that messages in given state will NOT be processed.
+     * Property is set for each state separately, in pattern:
+     * ASYNCH_FINAL_MESSAGES_PREFIX + <lower-case state name> + ASYNCH_FINAL_MESSAGES_SAVE_TIME_IN_SEC_SUFFIX
+     * Example:
+     * ohf.asynch.finalMessages.ok.saveTimeInSec
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_PREFIX = PREFIX + "asynch.finalMessages.";
+
+    /**
+     * Suffix to be used in conjuction with {@link CoreProps.ASYNCH_FINAL_MESSAGES_PREFIX}.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_SAVE_TIME_IN_SEC_SUFFIX = ".saveTimeInSec";
+
+    /**
      * Administrator email(s); if more emails, then separated them with semicolon, if empty then email won't be sent.
      */
     public static final String MAIL_ADMIN = PREFIX + "mail.admin";
--- a/core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/AbstractFinalMessageProcessor.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/AbstractFinalMessageProcessor.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.core.common.asynch.finalmessage;
+
+import javax.annotation.PostConstruct;
+
+import org.openhubframework.openhub.api.asynch.finalmessage.FinalMessageProcessor;
+import org.openhubframework.openhub.api.entity.Message;
+import org.openhubframework.openhub.core.common.dao.MessageDao;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.TransactionTemplate;
+
+/**
+ * Abstract class for custom implementations of {@link FinalMessageProcessor}.
+ *
+ * @author Karel Kovarik
+ * @since 2.1
+ */
+public abstract class AbstractFinalMessageProcessor implements FinalMessageProcessor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractFinalMessageProcessor.class);
+
+    @Autowired
+    protected MessageDao messageDao;
+
+    // transactionTemplate
+    @Autowired
+    protected TransactionTemplate transactionTemplate;
+
+    @Autowired
+    protected PlatformTransactionManager transactionManager;
+
+    @PostConstruct
+    void initializeTransactionManager() {
+        this.transactionTemplate = new TransactionTemplate(transactionManager);
+        this.transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRED);
+    }
+
+    @Override
+    public final void processMessage(final Message message) {
+        LOG.trace("Will process message with id [{}].", message.getId());
+
+        // execute in REQUIRED transaction
+        transactionTemplate.execute((TransactionStatus status) -> {
+            // fetch the message to have it in current hibernate session
+            final Message loaded = messageDao.findMessage(message.getId());
+
+            LOG.trace("Message fetched successfully");
+            // invoke the processing method.
+            doProcessMessage(loaded);
+            return null; // callback without result
+        });
+
+        LOG.trace("Message processing finished.");
+    }
+
+    /**
+     * Get instance of messageDao repository.
+     *
+     * @return the dao bean.
+     */
+    protected MessageDao getMessageDao() {
+        return messageDao;
+    }
+
+    /**
+     * Do process message.
+     *
+     * @param message the message entity.
+     */
+    protected abstract void doProcessMessage(Message message);
+}
--- a/core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.core.common.asynch.finalmessage;
+
+import org.openhubframework.openhub.api.asynch.finalmessage.FinalMessageProcessor;
+import org.openhubframework.openhub.api.entity.Message;
+import org.openhubframework.openhub.api.entity.Request;
+import org.openhubframework.openhub.core.common.dao.ExternalCallDao;
+import org.openhubframework.openhub.core.common.dao.RequestResponseDao;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+/**
+ * FinalMessageProcessor implementation that does delete given message from datastore.
+ *
+ * It does delete message and all related entities as well.
+ *
+ * @author Karel Kovarik
+ * @since 2.1
+ * @see FinalMessageProcessor
+ */
+@Qualifier(DeleteFinalMessageProcessor.QUALIFIER)
+public class DeleteFinalMessageProcessor extends AbstractFinalMessageProcessor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DeleteFinalMessageProcessor.class);
+
+    /**
+     * Qualifier used for the bean.
+     */
+    public static final String QUALIFIER = "deleteFinalMessageProcessor";
+
+    /**
+     * The order used in FinalMessageProcessor hierarchy, see {@link FinalMessageProcessor#getOrder()}.
+     */
+    public static final int ORDER = -1_000;
+
+    @Autowired
+    protected RequestResponseDao requestResponseDao;
+
+    @Autowired
+    protected ExternalCallDao externalCallDao;
+
+    @Override
+    protected void doProcessMessage(final Message message) {
+        // delete external calls
+        deleteExternalCalls(message);
+
+        // delete request responses
+        deleteRequestsResponses(message);
+
+        // delete message entity itself
+        deleteMessageEntity(message);
+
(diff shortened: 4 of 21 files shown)


core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
```diff
@@ -0,0 +1,146 @@
+/*
```
pjuza, line 1: add package-info.java files into packages ...

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
```diff
@@ -82,6 +82,37 @@
...
+    /**
+     * Maximum number of messages processed per job.
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_ITERATION_MESSAGE_LIMIT = PREFIX + "asynch.finalMessages.iterationMessageLimit";
+
+    /**
+     * Duration in seconds to keep messages in OK state in the datastore. After this period expires and
+     * there is no other action with the message, it will be processed as final (deleted probably).
+     */
+    public static final String ASYNCH_FINAL_MESSAGES_OK_SAVE_TIME_IN_SEC = PREFIX + "asynch.finalMessages.ok.saveTimeInSec";
```
pjuza: What does it mean if value is 0? Please describe it.
Maybe this comment is mainly for description on wiki to be clear what does it mean to use 0 value for time intervals.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @author Karel Kovarik
+ * @since 2.1
+ */
+public class FinalMessagesProcessingServiceImpl implements FinalMessagesProcessingService {
+    private static final Logger LOG = LoggerFactory.getLogger(FinalMessagesProcessingServiceImpl.class);
+
+    protected static final Map<MsgStateEnum, String> FINAL_MSG_STATES_CONFIG = new HashMap<>();
+    static {
+        FINAL_MSG_STATES_CONFIG.put(MsgStateEnum.OK, CoreProps.ASYNCH_FINAL_MESSAGES_OK_SAVE_TIME_IN_SEC);
```
pjuza: It would be nice to have definition of final states at one place only. See org.openhubframework.openhub.api.entity.MsgStateEnum#isFinal

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+        FINAL_MSG_STATES_CONFIG.put(MsgStateEnum.FAILED, CoreProps.ASYNCH_FINAL_MESSAGES_FAILED_SAVE_TIME_IN_SEC);
+        FINAL_MSG_STATES_CONFIG.put(MsgStateEnum.CANCEL, CoreProps.ASYNCH_FINAL_MESSAGES_CANCEL_SAVE_TIME_IN_SEC);
+    }
+
+    @ConfigurableValue(key = CoreProps.ASYNCH_FINAL_MESSAGES_ITERATION_MESSAGE_LIMIT)
+    protected ConfigurationItem<Long> messagesTechnicalLimit;
```
pjuza, line 74: I like using limit, good idea

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+        Assert.notNull(transactionManager, "the transactionManager must not be null");
+
+        this.transactionTemplate = new TransactionTemplate(transactionManager);
+        this.transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+    }
+
+    @Override
+    public void processMessage(final Message message) {
+        // notify listeners, note they are not wrapped in try catch,
+        // so if listener fails, it will effectivelly cancel the delete
```
pjuza: Mispelling - effectively

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    @Autowired
+    public DeleteFinalMessageProcessor(PlatformTransactionManager transactionManager) {
+        Assert.notNull(transactionManager, "the transactionManager must not be null");
+
+        this.transactionTemplate = new TransactionTemplate(transactionManager);
+        this.transactionTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+    }
+
+    @Override
+    public void processMessage(final Message message) {
```
pjuza: I had different idea how to implemented it - I wanted to define interface, similar to FinalMessageProcessor in your implementation. And this interface will have default implementation that deletes records from data store. And if you would like to change it for your project then you will implement it in your way. 
But your implementation deletes records in every case, only custom implementations before deletion itself are possible.

pjuza: I would add final modifier to this method - this method represents "core" of this class and should be unchangeble. You can override each delete method but core logic should be always same.

pjuza (changes requested):

core-api/src/main/java/org/openhubframework/openhub/api/asynch/finalmessage/FinalMessageProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.openhubframework.openhub.api.asynch.finalmessage;
+
+import org.openhubframework.openhub.api.entity.Message;
+
+/**
+ * Contract for final messages processor.
+ * Messages that are no-longer considered relevant for openhub async workflow are expected
+ * to be processed by this processor. After this, they should not be in OpenHub datastore
```
hanusto: According to comment from @pjuza it is wrong javadoc. _FinalMessageProcess_ is only about processing messages in final state. Removing or archiving is one of many action what can be done.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    protected void deleteMessageEntity(final Message message) {
+        messageDao.delete(message);
+    }
+
+    /**
+     * Get sorted list of listeners.
+     *
+     * @return list of listeners, that will be executed consequently.
+     */
+    protected List<DeleteFinalMessageListener> getSortedListenerList() {
```
hanusto: Why? If each component has Order annotation, it is sorted by default.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    protected ExternalCallDao externalCallDao;
+
+    // listeners
+    @Autowired(required = false)
+    protected Set<DeleteFinalMessageListener> deleteFinalMessageListenerSet = new HashSet<>();
+
+    protected List<DeleteFinalMessageListener> sortedListenerList;
+
+    @Autowired
+    public DeleteFinalMessageProcessor(PlatformTransactionManager transactionManager) {
```
hanusto: Which _ PlatformTransactionManager_? OpenHub?

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * Transaction:
+ * there is no need for open transaction, processor does start new and commits it.
+ *
+ * @author Karel Kovarik
+ * @since 2.1
+ * @see FinalMessageProcessor
+ * @see DeleteFinalMessageListener
+ */
+public class DeleteFinalMessageProcessor implements FinalMessageProcessor {
```
hanusto: It is not spring component?

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
hanusto: But we have to force to delete outdated messages. If this listener will be Spring component, it will be possible by Primary annotation but it should be exceptionally.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
hanusto: I see, this is safer. I agree.

core/src/main/resources/db/migration/postgresql/V1_0_3__final_messages.sql
```diff
@@ -0,0 +1,27 @@
+--
+-- core.async
+--
+
+-- final messages processing enabled or disabled.
```
hanusto: In future we should migrate from Flyway to Liquibase...

core/src/test/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImplTest.java
```diff
@@ -0,0 +1,208 @@
...
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import java.time.Instant;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.openhubframework.openhub.api.asynch.finalmessage.FinalMessageProcessor;
```
hanusto, line 30: In future we can reorder imports = it will be necessary to improve our code style.

hanusto (changes requested):

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
kkovarik: Ok, if you think it will ever by changed It´s small change as I was anticipating it :) 

I have to relative the relationship between State + Prop, so I will have to make dynamic resolving like:
`asynch.finalMessages.<lowercased-name-of-state>.saveTimeInSec`, ok?

What should happen if configuration value is not found, exception during startup? Other solution would be default value, but I´m not too fond of it.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
kkovarik: Fixed, thanks

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
kkovarik: Thanks, good point. If this is overriden, the very "core" will be changed.

Changed.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
kkovarik: Well, good question, was thinking about it as well. Relates to the PR https://github.com/OpenWiseSolutions/openhub-framework/pull/93 

It should be the transactionManager that manages openhub datasource. Which one is that? In proposition of https://github.com/OpenWiseSolutions/openhub-framework/pull/93 it is either Spring-Boot autoconfigured one (then only one is present), or the Primary bean, if multiple beans are present.

Hence wiring with the interface only, not with qualifier or anything.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
kkovarik: It is, however it is registered in Configuration class - see FinalMessageProcessingConfiguration, motivation was to register it only if this whole functionality is enabled

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/DeleteFinalMessageProcessor.java
kkovarik: @pjuza  thanks for mentioning it, DeleteFinalMessageProcessor is now ConditionalOnMissing bean, meaning it can be overriden easily, not just with Primary bean. Then you can simple achieve different behaviour with your custom FinalMessageProcessor.

However I think in most cases, custom implementation will want to delete the messages ultimately. Just implement custom "archiving" before that. That is achievable with the listener only. Listener has some control as well, meaning, if it throws exception, delete will not continue. That way it could hopefully ensure, that messages won´t be lost unexpectedly.

core-api/src/main/java/org/openhubframework/openhub/api/asynch/finalmessage/FinalMessageProcessor.java
kkovarik: Thanks, good point, updated javadoc.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
kkovarik: Thanks, added

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
kkovarik: Setting to '0' means, that messages will be processed as soon as possible. Effectively meaning next scheduled job run,

Javadoc improved, thanks good idea to pointing that out.

pjuza: And what value should I use if I want to have specific messages forever in data store?

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
pjuza: I would throw exception during startup ...

kkovarik:
Updated implementation & description. Now there is job, service and collection of FinalMessageProcessors. One of the MessageProcessor is DeleteFinalMessageProcessor. Please check PR again.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingJob.java
```diff
@@ -0,0 +1,64 @@
...
+    private FinalMessagesProcessingService finalMessagesProcessingService;
+
+    @OpenHubQuartzJob(
+            name = JOB_NAME,
+            executeTypeInCluster = JobExecuteTypeInCluster.NOT_CONCURRENT,
+            simpleTriggers = @QuartzSimpleTrigger(
+                    repeatIntervalProperty = CoreProps.ASYNCH_FINAL_MESSAGES_PROCESSING_INTERVAL_SEC,
+                    intervalPropertyUnit = SimpleTriggerPropertyUnit.SECONDS)
+    )
+    public final void processFinalMessages() {
```
pjuza, line 55: I would appreacite if I could disable job at all. Maybe also with some configuration parameter ...
I would use the same parameter you already have: org.openhubframework.openhub.api.configuration.CoreProps#ASYNCH_FINAL_MESSAGES_PROCESSING_ENABLED

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
```diff
@@ -0,0 +1,193 @@
...
+                        // invoke all the processors
+                        for (FinalMessageProcessor finalMessageProcessor : finalMessageProcessorList) {
+                            finalMessageProcessor.processMessage(message);
+                        }
+                       return null; // callback without result
+                    });
+
+                } catch (Exception ex) {
+                    // continue with another message if handling of one fails.
+                    LOG.warn("Failed to process message : {}, will continue with next.", message, ex);
```
pjuza: I think it's "normal" error => please change log level to error.

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/AbstractFinalMessageProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+ * @since 2.1
+ */
+public abstract class AbstractFinalMessageProcessor implements FinalMessageProcessor {
+
+    @Autowired
+    protected MessageDao messageDao;
+
+    @Transactional(propagation = Propagation.REQUIRED)
+    @Override
+    public void processMessage(final Message message) {
```
pjuza: this method should be final

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/AbstractFinalMessageProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+        // fetch the message to have it in current hibernate session
+        final Message loaded = messageDao.findMessage(message.getId());
+
```
pjuza, line 66: I would add logging on level DEBUG/TRACE ...

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessageProcessingConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.openhubframework.openhub.core.common.asynch.finalmessage;
+
+import org.openhubframework.openhub.api.asynch.finalmessage.FinalMessageProcessor;
+import org.openhubframework.openhub.api.asynch.finalmessage.FinalMessagesProcessingService;
+import org.openhubframework.openhub.api.configuration.CoreProps;
+import org.springframework.beans.factory.annotation.Qualifier;
```
pjuza: Unused imports ...

pjuza (approved):

pjuza:
Please dont't forget to add new wiki page - I prefer to add new page under 
https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33573/User+guide. It's quite new feature :)

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingJob.java
kkovarik: It is already implemented that way. Job is not on packageScan, it is registered as bean conditionaly in the configuration class **org.openhubframework.openhub.core.common.asynch.finalmessage.FinalMessageProcessingConfiguration**

idea is that none if this beans (job, service, processor) is registered, if the property  ASYNCH_FINAL_MESSAGES_PROCESSING_ENABLED is false

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessageProcessingConfiguration.java
kkovarik: Thanks!

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/FinalMessagesProcessingServiceImpl.java
kkovarik: Changed

core/src/main/java/org/openhubframework/openhub/core/common/asynch/finalmessage/AbstractFinalMessageProcessor.java
kkovarik: Thanks for mentioning it, added logging to AbstractFinalMessageProcessor and added seting MDC Log context in FinalMessageProcessingService, so that logs of message processing are correlated via correlationId.

kkovarik:
Added documentation https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/778764295/Messages+in+final+state and configuration described in https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33599/OpenHub+configuration

core-api/src/main/java/org/openhubframework/openhub/api/configuration/CoreProps.java
kkovarik: Added possibility to set value to "-1", then it will be kept indefinitely

pjuza:
My last two comments to AbstractFinalMessageProcessor - I would change logging level to trace from debug because often there is huge amount of messages and I would add add final to processMessage method.
I like your solution!

kkovarik:
Log level changed & processMessage not final

hanusto (approved):

PR #95, opened 2018-11-14
[OHFJIRA-100] : CloseableHttpComponentsMessageSender moved to api module

kkovarik:
#### Issue
* in wiki, it is described that openhub-framework provides Http Message Sender : https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33674/HTTP+Message+Sender
However now it is in core module

#### Changes
* moved class to the core-api module

#### Discussion
* it is breaking change, however it is not in api module. Therefore will it be ok, to just move it & describe in release notes? or should we go with Deprecation and duplicating class till next major release?

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -51,6 +51,10 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-autoconfigure</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.ws</groupId>
+            <artifactId>spring-ws-core</artifactId>
+        </dependency>
         <!-- enabled configuration properties in OHF classes -->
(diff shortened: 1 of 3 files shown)


pjuza:
I'm thinking if it's not better to move it to common module. Maybe it's only my "feeling" but this class is not really api class, is it? What do you think?

I would "just move it & describe in release notes"

pjuza (approved):

kkovarik:
Moved from core to common module.

Implementation is kept intact, for migration only package name should be changed. Mentioned in Road planner, wiki documentation https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33674/HTTP+Message+Sender

PR #96, opened 2018-11-15
[OHFJIRA-96] : Camel core version updated to 2.21.3

kkovarik:
## Upgrade to Camel core 2.21.3

* it is the latest version (2.21.x) supporting Spring Boot 1.5.x

### Changes needed
* Order of camel´s RoutesCollector has changed: added Order for **ApplicationContextsRegistry**, as it initialises itself on ContextRefreshedEvent. It needs to receive the event before **RoutesCollector**, otherwise camel context **won´t start**, as registry is empty
* FatJarRouter : removed entirely, see: https://stackoverflow.com/questions/44723205/using-fatjarrouter-when-upgrading-camel-spring-boot-from-2-18-to-2-19 , therefore removed the handling.
* Quartz: quartz Scheduler interface has new method (quartz 2.2.3 -> quartz 2.3.0), added implementation. other than that quartz update should be about bugfixes, https://github.com/quartz-scheduler/quartz/releases/tag/quartz-2.3.0, database schema is kept intact

### Issue to resolve
Two unit tests are now failing. It is test of RequestResponseTest:
testSavingResponseOnly
testSavingResponseWithMsgOnly

Meaning - saving response without request (only ResponseReceiveEventNotifier is active, RequestSendingEventNotifier is NOT).

Reason is this optimalization in Camel: https://issues.apache.org/jira/browse/CAMEL-11346 which made some changes in camel 2.20.x+, see this commit https://github.com/apache/camel/commit/fdfd57d47684ef575f2dc15b9309978aa071890c
and changes in SendProducer class. It effectivelly means, that ExchangeSentEvent is emitted only if EventHelper.notifyExchangeSending(...) returns true = meaning there is listener for ExchangeSendingEvent. It means, that ResponseReceiveEventNotifier does not work properly without RequestSendingEventNotifier enabled.

Do we still want to support this scenario - response only saving? If yes, I am afraid it will lead to some "hackish" solution like that Response is pretending to be consumer of ExchangeSendingEvent as well, or to have regular support for "two event" listeners.

### DIFF
--- a/core/src/main/java/org/openhubframework/openhub/core/common/camel/ApplicationContextsRegistry.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/common/camel/ApplicationContextsRegistry.java
@@ -33,6 +33,7 @@ import org.springframework.context.ApplicationListener;
 import org.springframework.context.event.ApplicationContextEvent;
 import org.springframework.context.event.ContextRefreshedEvent;
 import org.springframework.context.event.ContextStartedEvent;
+import org.springframework.core.Ordered;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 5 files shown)


pjuza (approved):

pjuza:
I have checked dependency tree - it's ok. Also build on my local computer is ok.

> Do we still want to support this scenario - response only saving?

No and it's wasn't never supported. If you look at org.openhubframework.openhub.core.reqres.ResponseReceiveEventNotifier you will see that it's possible to have requests without responses but not vice verce.

pjuza:
If anything change with regard to request/response then please change also wiki - https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33606/Request+response+tracking

PR #111, opened 2020-04-07
[NOJIRA] : MessageServiceImpl nested transactions

kkovarik:
Change of transaction propagation in MessageServiceImpl.

insertMessage and setStateInQueue do start new transaction always

Please review if it makes sense, as it was changed as "hotfix" :)


hanusto:
It's long time PR. Now we should probably reject this PR.

kkovarik:
Closing, not to change current behaviour

PR #112, opened 2020-04-08
[OHFJIRA-107] : fixed final message processing of messages without re…

sabolm:
### Issue

- when final message processing job is turned on
- saving requests and responses is turned on
- output message is without response (messaging is used)
- message is not deleted with error: `the response must not be null.; nested exception is java.lang.IllegalArgumentException: the response must not be null.`

### Changes
- added check for empty response


kkovarik (approved):
👍

PR #114, opened 2020-04-24
[OHFJIRA-109] : changed selection of partly failed and postponed mess…

sabolm:
### Current
MessagesPool finds messages in the PARTLY_FAILED and POSTPONED state for further processing. Currently all postponed messages are processed first and then are processed all partly failed messages. 

This behaviour has negative performance impact in combination with messages in guaranteed order. If there are many POSTPONED messages and the oldest message is in PARTLY_FAILED state then MessagesPool has to go through all POSTPONED messages but they can't be processed. 

### Improvement
This improvement combines selection of these two message groups into common SQL query.

### DIFF
--- a/core/src/main/java/org/openhubframework/openhub/core/common/dao/MessageDao.java
+++ b/core/src/main/java/org/openhubframework/openhub/core/common/dao/MessageDao.java
@@ -126,6 +126,16 @@ public interface MessageDao {
     Message findPostponedMessage(Duration interval);
 
     /**
+     * Finds ONE message in state {@link MsgStateEnum#POSTPONED} or {@link MsgStateEnum#PARTLY_FAILED}
+     *
+     * @param postponedInterval Interval (in seconds) after that can be postponed message processed again
+     * @param partiallyFailedInterval Interval (in seconds) after that can be partly failed message processed again
+     * @return message or null if there is no any message
+     */
+    @Nullable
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/openhubframework/openhub/core/common/dao/MessageDao.java
```diff
@@ -125,6 +125,16 @@
...
 
+    /**
+     * Finds ONE message in state {@link MsgStateEnum#POSTPONED} or {@link MsgStateEnum#PARTLY_FAILED}
+     *
+     * @param postponedInterval Interval (in seconds) after that can be postponed message processed again
+     * @param partiallyFailedInterval Interval (in seconds) after that can be partly failed message processed again
+     * @return message or null if there is no any message
+     */
+    @Nullable
+    Message findPostponedOrPartlyFailedMessage(Duration postponedInterval, Duration partiallyFailedInterval);
```
kkovarik: Litte inconsistency in naming - partlyFailedMessages vs partiallyFailedInterval?

kkovarik (approved):
👍 Makes sense to me

core/src/main/java/org/openhubframework/openhub/core/common/dao/MessageDao.java
sabolm: Thx, fixed to partlyFailedInterval

PR #115, opened 2020-04-27
[OHFJIRA-102] : added new configuration parameter for enabling admin emails

sabolm:
### ISSUE
When message fails then email is sent to recipients defined by `ohf.mail.admin` parameter. If parameter is empty then email won't be sent. So it's not clear how to enable or disable this feature.

### CHANGES
Add new parameter `ohf.mail.admin.enabled` for switching on/off for sending admin emails.
Added parameter to default DB configuration with value `false`.

### DOCUMENTATION
Documentation needs to be extended on page  https://openhubframework.atlassian.net/wiki/spaces/OHF/pages/33599/OpenHub+configuration


kkovarik (approved):
Seems alright, thanks :)

PR #116, opened 2020-04-28
[OHFJIRA-110] : guaranteed order messages query optimization

sabolm:
### ISSUE
There is dao `getMessagesForGuaranteedOrderForRoute` which returns list of all messages which fits the query conditions (there isn't limit). This has negative performance impact (memory, time) for messages in guaranteed order. 

### CHANGES
- added new method `getMessagesForGuaranteedOrderForRoute` with `limit` to MessageService and MessageDAO
- deprecated existing `getMessagesForGuaranteedOrderForRoute` methods without limit
- switched all deprecated methods with a new ones

### DIFF
--- a/core-spi/src/main/java/org/openhubframework/openhub/spi/msg/MessageService.java
+++ b/core-spi/src/main/java/org/openhubframework/openhub/spi/msg/MessageService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2014-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


core-spi/src/main/java/org/openhubframework/openhub/spi/msg/MessageService.java
```diff
@@ -209,9 +209,22 @@ void setStateFailed(@Header(MSG_HEADER) Message msg,
      * @param excludeFailedState {@link MsgStateEnum#FAILED FAILED} state is used by default;
      *                           use {@code true} if you want to exclude FAILED state
      * @return list of messages ordered by {@link Message#getMsgTimestamp() message timestamp}
+     * @deprecated use {@link #getMessagesForGuaranteedOrderForRoute(String, boolean, long)} instead
```
kkovarik: How about to explicit mention that this method does return all of the messages, and thats the reason why to use the overloaded variant? To avoid any possible confusion :)

core/src/main/java/org/openhubframework/openhub/core/common/asynch/AsynchInMessageRoute.java
```diff
@@ -85,6 +85,8 @@
 
     static final String ROUTE_ID_GUARANTEED_ORDER = "guaranteedOrder" + AbstractBasicRoute.ROUTE_SUFFIX;
 
+    private static final long GUARANTEED_ORDER_MESSAGES_LIMIT = 2L;
```
kkovarik, line 88: 👍

kkovarik (approved):
Makes sense to me. 👍

core-spi/src/main/java/org/openhubframework/openhub/spi/msg/MessageService.java
sabolm: Good point. I'll improve documentation.

PR #118, opened 2020-05-19
[OHFJIRA-113] : fixed DB throttling disable configuration property

sabolm:
### ISSUE
Changing property `ohf.throttling.disable` in DB hasn't got any effect on enabling/disabling throttling.
Implementation expects property name `ohf.disable.throttling`.

### CHANGES
- changed property name in H2 and Postgresql scripts to `ohf.disable.throttling`
- fixed fe mocks


kkovarik (approved):
Does make sense, thanks

